PR #2, opened 2017-12-15
Avoid kubeclient issue

jhernand:
The `kubeclient` gem does not provide exactly the same kind of objects inside notices and inside query responses. That should be fixed by abonas/kubeclient#279. Meanwhile this pull request introduces patches in the provider to avoid the issue.

The first  patch can and should be reverted when the issue in `kubeclient` is fixed. The second one should stay.


jhernand:
@masayag please review.

PR #3, opened 2017-12-15
Add missing `Clone to VM` dialog

jhernand:
The provider already supports provisioning of virtual machines from
templates, but the dialog is missing. This patch adds it.

Request tab:

kubevirt-request

Purpose tab:

kubevirt-purpose

Catalog tab:

kubevirt-catalog

Hardware tab:

kubevirt-hardware

Schedule tab:

kubevirt-schedule

### DIFF
--- a/content/miq_dialogs/miq_provision_kubevirt_dialogs_clone_to_vm.yaml
+++ b/content/miq_dialogs/miq_provision_kubevirt_dialogs_clone_to_vm.yaml
@@ -1,0 +1,378 @@
+---
+:name: miq_provision_kubevirt_dialogs_clone_to_vm
+:description: Create virtual machine from template
+:dialog_type: MiqProvisionWorkflow
+:content:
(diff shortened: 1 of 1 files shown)


jhernand:
@masayag please review.

masayag:
@jhernand, please include also screenshots of the different dialog tabs.
In addition, isn't there a meaning for the 'Environment' tab in this context ?
Can't user specify the desired cluster where to run the vm for instance ?

jhernand:
@masayag there is only one _cluster_ in KubeVirt, so it makes no sense to let the user select it, it will always be the same. I am not aware of any other _environment_ thing that makes sense in the current KubeVirt architecture, that is why it isn't part of the dialog. I think we can easily add it later if needed.

I will include the screenshots.

jhernand:
@masayag I added the screen shots of the tabs of the dialog.

PR #4, opened 2017-12-15
Adapt the provider to the latest KubeVirt API

jhernand:
The KubeVirt team has decided to use `OfflineVirtualMachine` as the name
for the custom resource that represents virtual machines that are not
running. We assumed `StoredVirtualMachine` in the past. This patch
changes the provider to use the new name and to unify the names used for
methods and variables. They will be now like this:

  OfflineVirtualMachine -> offline_vm
  VirtualMachine -> live_vm
  Template -> template


jhernand:
@masayag  please review.

app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -111,18 +111,56 @@ def process_node(object)
...
+  def process_live_vm(object)
     # Get the basic information:
     uid = object.metadata.uid
     name = object.metadata.name
     domain = object.spec.domain
 
+    # Get the identifier of the offline virtual machine from the owner reference:
+    owner = find_owner(object, 'OfflineVirtualMachine')
+    unless owner
+      _log.info(
```
masayag: this is not recognized here - you should add:

`  include Vmdb::Logging`

jhernand: Thanks @masayag, did that in the latest version of the pull request.

app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/power.rb
```diff
@@ -17,23 +17,49 @@
...
+      connection.create_live_vm(live_vm)
     end
   end
 
   def raw_stop
     ext_management_system.with_provider_connection do |connection|
+      # Retrieve the details of the offline virtual machine:
+      offline_vm = connection.offline_vm(name)
+
+      # Change the `running` attribute to `false` so tha the offline virtual machine controller will take it and delete
```
masayag: s/tha/that

PR #6, opened 2017-12-28
KubeVirt as a secondary manager of Container Providers

masayag:
The KubeVirt provider will be managed as a secondary provider of
OpenShift/Kubernetes providers.

The endpoint and authentication of kubevirt will reside under the
container providers, selected by the :kubevirt role and authkey.

Adding and removal of KubeVirt should be performed from the container
manager virtualization tab.

Depends on https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/197

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager.rb
@@ -23,6 +23,13 @@ class ManageIQ::Providers::Kubevirt::InfraManager < ManageIQ::Providers::InfraMa
   require_nested :Template
   require_nested :Vm
 
+  include ManageIQ::Providers::Kubernetes::VirtualizationManagerMixin
+
+  belongs_to :parent_manager,
+             :foreign_key => :parent_ems_id,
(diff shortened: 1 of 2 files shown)


masayag:
[add detect logic to openshift provider - large: 2 weeks](https://trello.com/c/H7iH4QqQ/12-add-detect-logic-to-openshift-provider-large-2-weeks)

app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -23,6 +23,13 @@ class ManageIQ::Providers::Kubevirt::InfraManager < ManageIQ::Providers::InfraMa
   require_nested :Template
   require_nested :Vm
 
+  include ManageIQ::Providers::Kubernetes::VirtualizationManagerMixin
```
masayag: this one adds dependency on the manageiq-providers-kubernetes project, therefore we add an explicit dependency on it

Fryguy:
@masayag Please do not merge your own PRs...that is a general ManageIQ policy...someone else should review and merge.

PR #8, opened 2018-01-10
create a virtual machine template user guide

yanirq:
Documentation describing how to write a Virtual machine template for ManageIQ

### DIFF
--- a/docs/virtual_machine_template_guide.adoc
+++ b/docs/virtual_machine_template_guide.adoc
@@ -1,0 +1,222 @@
+= Virtual Machine Templates
+
+Virtual machine templates are openshift templates (link:https://docs.openshift.org/latest/dev_guide/templates.html[Openshift Templates]) +
+that describe virtual machine specifications which produce the following template objects: +
+**One** OfflineVirtualMachine and **0-N** PersistentVolumeClaims. +
+The following document describes the process of creating a Virtual machine template which will be managed by ManageIQ.
+
+== Writing templates
+You can define new virtual machine templates to make it easy to recreate all relevant objects (OfflineVirtualMachine and PersistentVolumeClaim). The template will define the objects it creates along with some metadata to guide the creation of those objects.
+
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha1
+kind: Template
+metadata:
+  name: RHEL-7 ❶
+  annotations:
+  labels:
+    miq.github.io/kubevirt-os: rhel-7 ❷
+    miq.github.io/kubevirt-is-vm-template: True ❸
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine ❹
+  metadata:
+    name: ${NAME} ➎
+  spec:
+    template:
+      metadata:
+        labels:
+          my: label
+      spec:
+        domain: 
+          resources: ➏
+            requests:
+              memory: ${MEMORY} 
+          cpu: ❼
+            cores: ${CPU}
+          devices:
+            disks: ❽
+              name: disk0
+              volumeName: mypvc 
+      volumes: ➒
+       - name: mypvc
+         presistentVolumeClaim:
+           claimName: rhel-7-pvc-${NAME}
+
+- apiVersion: v1
+  kind: PersistentVolumeClaim ➓
+  metadata:
+    name: rhel-7-pvc-${NAME} ⓫
+    #FIXME: add an annotation to specify which volume this pvc should be based on
+  spec:
+    accessModes:
+    - ReadWriteOnce
+    resources:
+      requests:
+        storage: 1Gi 
+parameters: ...
+
+----
+ 
+❶ The unique name of the template. +
+
+❷ OS identifier of the template. +
+
+❸ Indicator if this is a kubevirt template. +
+
+❹ The OfflineVirtualMachine Object. This object must appear only once in the template. +
+
+➎ OfflineVirtualMachine’s name. +
+
+➏ OfflineVirtualMachine’s resources, currently includes under requirements only memory. +
+   Memory is the minimal amount the Virtual Machine requires. +
+   These attributes are represented as parameters that will either be filled out by the user Or receive default values. +
+   
+❼ Number of cpu cores for the Virtual Machine. +
+  This attribute is represented as parameter that will either be filled out by the user Or receive a default value. +
+  
+❽ Disk contains a "volumeName" attribute, which links it to a volume +
+
+➒ Specification of a disk's volume from the disks section. It contains persistent volume claims.  
+(more about link:https://kubernetes.io/docs/concepts/storage/persistent-volumes/[Persistent Volume Claims] ) +
+        
+➓ The PersistenVolumeClaim Object that accompanies the OfflineVirtualMachine Object. +
+
+⓫ PersistenVolumeClaims should include: name, permissions and requested amount of storage. +
+
+
+
+=== Labels
+
+Labels will be presented on every OCP virtual machine template in order for ManageIQ to recognize the KubeVirt template characteristics. + 
+The label "miq.github.io/kubevirt-os" is type of OS the template refers to. +
+The label "miq.github.io/kubevirt-is-vm-template" will indicate with a boolean value if it is a virtual machine template.
+
+[source,yaml]
+----
+kind: Template
+metadata:
+  name: RHEL-7
+  labels:
+    miq.github.io/kubevirt-os: rhel-7 
+    miq.github.io/kubevirt-is-vm-template: true
+----
+
+=== Parameters
+
+Parameters allow a value to be supplied by the user or generated when the template is instantiated. +
+Then, that value is substituted wherever the parameter is referenced. +
+The following parameters will be used for the OfflineVirtualMachine name (and as a part of the persistent volume claim names) +
+Default values will be provided for memory and cpu, which will be used if the user does not supply a different value.
+
+[source,yaml]
+----
+parameters:
+- description: Name for the new VM     
+  name: NAME   
+- name: MEMORY
+  description: Amount of memory
+  value: 512Mi  
+- name: CPU
+  description: Amount of cores
+  value: 4    
+----
+
+Additional info on link:https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/[compute resources]
+
+
+== Creating the template object in the cluster
+
+If you have a YAML file that defines a virtual machine template, you can upload the template to an Openshift cluster using the CLI. 
+To upload a template to your current project’s template library, pass the YAML file with the following command: +
+
+----
+$ oc create -f <filename>
+----
+
+
+=== Creating virtual machine template using the Web Console
+TBD  (add screenshots)
+
+=== Generating virtual machine template from CLI 
+The list of parameters that you can override are listed under  Parameters.
+You can list them with the CLI by using the following command and specifying the file to be used:
+----
+  $ oc process --parameters -f <filename>
+----
+
+Or create objects from a template by processing the template and piping the output to oc create:
+
+----
+  $ oc process --parameters -f <filename> | oc create -f -
+----
+
+
+== Modifying an uploaded template
+You can edit a template that has already been uploaded to your project by using the following command: +
+
+----
+$ oc edit template <template>
+----
+
+
+
+== Full virtual machine template example
+
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha1
+kind: Template 
+metadata:
+  name: RHEL-7 
+  # name: Microsoft-Windows-2012r1
+  labels:
+    miq.github.io/kubevirt-os: rhel-7
+    miq.github.io/kubevirt-is-vm-template: true  
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine
+  metadata:
+    name: ${NAME} 
+  spec:
+    template:
+      metadata:
+        labels:
+          my: label
+      spec:
+        domain:
+          resources:
+            requests:
+              memory: ${MEMORY} 
+          cpu:
+            cores: ${CPU}
+          devices:
+            disks:
+              name: disk0
+              volumeName: mypvc 
+      volumes:
+       - name: mypvc
+         presistentVolumeClaim:
+           claimName: rhel-7-pvc-${NAME}
+- apiVersion: v1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: rhel-7-pvc-${NAME}
+    #FIXME: add an annotation to specify which volume this pvc should be based on
+  spec:
+    accessModes:
+    - ReadWriteOnce
+    resources:
+      requests:
+        storage: 1Gi
+parameters:
+- description: Name for the new VM     
+  name: NAME   
+- description: Amount of memory
+  name: MEMORY
+  value: 512Mi  #default
+- description: Amount of cores
+  name: CPU
+  value: 4    #default
+----


docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+  metadata:
+    name: rhel-7-${NAME} ❸
+  spec:
+    template:
+      metadata:
+      labels:
+        my: label
+      spec:
+        domain:
+          resources: ❹
```
rmohr: The `requests` is missing:

```yaml
    resources:
      requests:
        memory: ${MEMORY}
        cpu: ${CPU}
```

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+  template: “miq.github.io/kubvirt/vm-template”           
+parameters:
+- description: Name for the new VM     
+  name: NAME   
+- description: Amount of memory in MB
+  name: MEMORY
+ 
+  value: 512Mi  #default
+- description: Amount of cores
+  name: CPU
```
rmohr: `cpu` in the `resources.requests` section refers to how much resources you want to assign to the vm (cgroup related like in https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/). `spec.domain.cpu.cores` specifies the number of cpu cores in the vm (see https://github.com/kubevirt/kubevirt/pull/638).

I wonder what the cpu template variable in the other virtualization providers in ManageIQ refers to. I would expect that people are more interested in dynamically specifying the resource allocation than the actual number of cpu cores. @jhernand maybe you know more about that?

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+apiVersion: kubevirt.io/v1alpha1
+kind: Template 
+metadata:
+  name: RHEL-7  #This will define the template’s name
+  # name: Microsoft-Windows-2012r1
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: offlineVirtualMachine
+  metadata:
+    name: rhel-7-${NAME}
```
rmohr: It does not matter since every user can use her own preferred pattern, but maybe adding an `os: rhel-7` label and having a not-prefixed name mapping for the example might look nicer.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha1
+kind: Template
+metadata:
+  name: RHEL-7 ❶
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: offlineVirtualMachine ❷
```
rmohr: `OfflineVirtualMachine`

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha1
+kind: Template 
+metadata:
+  name: RHEL-7  #This will define the template’s name
+  # name: Microsoft-Windows-2012r1
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: offlineVirtualMachine
```
rmohr: `OfflineVirtualMachine`

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+----
+labels:
+  # vm-template label applied to all objects created from this template
+  template: “miq.github.io/kubvirt/vm-template”
+----
+
+There is also the ability to add labels in the template from the command line:
+
+----
+$ oc process -f <filename> -l name=“miq.github.io/kubvirt/vm-template”
```
rmohr: All the quote signs look like they are a non-default ascii character.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+----
+
+=== Parameters
+
+Parameters allow a value to be supplied by the user or generated when the template is instantiated. +
```
rmohr: I see `+` artifacts here.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+- apiVersion: kubevirt.io/v1alpha1
+  kind: presistentVolumeClaim ❼
+  metadata:
+    name: rhel-7-pvc-${NAME} ❽
+labels: ...       
+parameters: ...
+
+----
+ 
+❶ The unique name of the template. +
```
rmohr: I normally tend to not use such special characters in markdown, but the result looks nice.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+----
+
+Writing Templates
+You can define new VirtualMachine templates to make it easy to recreate all relevant objects (offlineVirtualMachine and PersistentVolumeClaim). The template will define the objects it creates along with some metadata to guide the creation of those objects.
+
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha1
+kind: Template
+metadata:
```
rmohr: I think we were also talking about a special ManageIQ label so that ManageIQ will pick up the template.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
+= VirtualMachine Templates
+
+VirtualMachine templates are openshift templates (https://docs.openshift.org/latest/dev_guide/templates.html) +
+that describe VirtualMachine specifications which produce the following template objects: +
+**One** OfflineVirtualMachine and PersistentVolumeClaim. +
```
pkliczewski: If we stated how many offline vms is produced I think we should add cardinality for pvc(s) as well.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+The following document describes the process of creating a Virtual machine template which will be managed by ManageIQ.
+
+== Uploading a Template
+
+If you have a YAML file that defines a VirtualMachine template, you can upload the template to an Openshift cluster using the CLI.
```
pkliczewski: Why to limit ourselves to specific way of uploading method?

docs/virtual_machine_template_guide.adoc
pkliczewski: At what point we would benefit from this specific label? With my limited understanding I think we need a way to be able to ask for VirtualMachine templates only.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,189 @@
...
+ 
+❶ The unique name of the template. +
+
+❷ The OfflineVirtualMachine Object. This object must appear only once in the template. +
+
+❸ OfflineVirtualMachine’s name. +
+
+❹ OfflineVirtualMachine’s resources, currently includes only memory and cpu. +
+   Cpu is the number of cpu cores the Virtual Machine requires. +
+   Memory is the minimal amount the Virtual Machine requires. +
```
pkliczewski: Can we add information how units are encoded?

docs/virtual_machine_template_guide.adoc
pkliczewski: [Here](https://pemcg.gitbooks.io/introduction-to-cloudforms-automation/content/chapter15/provisioning_dialog.html) you can see how hardware section looks like in vm provisioning dialog.

docs/virtual_machine_template_guide.adoc
rmohr: @yanirq apart from examples in the k8s docs, here is the best description i know: https://github.com/kubernetes/kubernetes/blob/a523084d88edc929d817f828d66b3091350d2f17/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity.go#L35

docs/virtual_machine_template_guide.adoc
rmohr: @pkliczewski If I remember it correctly, it was decided in an offline discussion, that ManageIQ will filter the templates based on such a label. You would simply be able to just ask for templates with that label attached to the REST call. You would still have to validate the template (e.g. that an OfflineVirtualMachine is present, that only supported parameters are there, ...).

docs/virtual_machine_template_guide.adoc
yanirq: these are new line characters for the adoc , is there a better alternative ?

docs/virtual_machine_template_guide.adoc
yanirq: acked

docs/virtual_machine_template_guide.adoc
pkliczewski: Thanks for clarification

docs/virtual_machine_template_guide.adoc
yanirq: I think so too,
@fabiand  do we want to restrict the PVC to one currently ?   or mention a 1-N cardinality here ?

docs/virtual_machine_template_guide.adoc
yanirq: Are we talking about CLI or file types ? 
We can add other upload methods here later on

docs/virtual_machine_template_guide.adoc
yanirq: Thats what the labels section in the document is for,
The labels are at the bottom of the file on purpose, and in the form of:

labels:
template: “miq.github.io/kubvirt/vm-template”

“miq.github.io/kubvirt/vm-template” label applied to all objects created from this template
and without having a flag of true/flase

the syntax is mentioned here:
https://docs.openshift.org/latest/dev_guide/templates.html#writing-labels

If that dosent answer the needs, let me know

docs/virtual_machine_template_guide.adoc
yanirq: It resembles openshift guides, if it will get lost in translation in the future then ill change it.

docs/virtual_machine_template_guide.adoc
yanirq: ack

docs/virtual_machine_template_guide.adoc
yanirq: might be right,  we have the same discussion in #9

docs/virtual_machine_template_guide.adoc
yanirq: ack, changed + added cores as well.

docs/virtual_machine_template_guide.adoc
rmohr: > The labels are at the bottom of the file on purpose, and in the form of:

Ah ok. That does not help filtering the template themselves. You will need the label in the templates label section `^metadata.labels` to query templates for it.

docs/virtual_machine_template_guide.adoc
rmohr: I managed to ignore that it is adoc and not markdown :+1:

docs/virtual_machine_template_guide.adoc
rmohr: since we can't specify templates (e.g. size) for a single PVC, we can allow adding an arbitrary number.

docs/virtual_machine_template_guide.adoc
yanirq: ok, 
do we still need to label the objects as mentioned above or this section can be removed ?

docs/virtual_machine_template_guide.adoc
yanirq: added a link

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
+= VirtualMachine Templates
+
+VirtualMachine templates are openshift templates (https://docs.openshift.org/latest/dev_guide/templates.html) +
+that describe VirtualMachine specifications which produce the following template objects: +
+**One** OfflineVirtualMachine and **1-N** PersistentVolumeClaims. +
+The following document describes the process of creating a Virtual machine template which will be managed by ManageIQ.
+
+== Uploading a Template
```
fabiand: Maybe call it "Creating the Template object in the cluster"?

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+----
+
+== Modifying an Uploaded Template
+You can edit a template that has already been uploaded to your project by using the following command: +
+
+----
+$ oc edit template <template>
+----
+
+Writing Templates
```
fabiand: Shouldn't this be a title i.e. "== Writing Tempates"

Also: What about changing the worder to:
1. writing
2. uploading
3. edit uploaded

docs/virtual_machine_template_guide.adoc
fabiand: The label must be on the metadata of the template - but that is already the case in line 34. (See below)

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+You can define new VirtualMachine templates to make it easy to recreate all relevant objects (OfflineVirtualMachine and PersistentVolumeClaim). The template will define the objects it creates along with some metadata to guide the creation of those objects.
+
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha1
+kind: Template
+metadata:
+  name: RHEL-7 ❶
+  labels:
+    miq.github.io/kubvirt/is-vm-template: True
```
fabiand: Here is the label.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine ❷
+  metadata:
+    name: rhel-7-${NAME} ❸
+  spec:
+    template:
+      metadata:
+      labels:
```
fabiand: labels are a subfield in metadata, please indent accordingly.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+        my: label
+      spec:
+        domain:
+          resources: ❹
+            requests:
+              memory: ${MEMORY} 
+              cpu:
+                cores: ${CPU}
```
fabiand: Cores are elsewhere, the layout is:
```yaml
spec:
  domain:
    resources:
      requests:
        memory: 4
    cpu:
      cores: 2
```

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+          devices:
+      disks: ➎
```
fabiand: disks is under devices, thus it has to be indented

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+        name: disk0
+        volumeName: mypvc 
+      volumes: ➏
+         name: mypvc
+         presistentVolumeClaim:
+      claimName: rhel-7-pvc-${NAME}
+- apiVersion: kubevirt.io/v1alpha1
+  kind: presistentVolumeClaim ❼
```
fabiand: Typoes in here.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+              cpu:
+                cores: ${CPU}
+          devices:
+      disks: ➎
+        name: disk0
+        volumeName: mypvc 
+      volumes: ➏
+         name: mypvc
+         presistentVolumeClaim:
+      claimName: rhel-7-pvc-${NAME}
```
fabiand: indetation needs to be fixed

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+            requests:
+              memory: ${MEMORY} 
+              cpu:
+                cores: ${CPU}
+          devices:
+      disks: ➎
+        name: disk0
+        volumeName: mypvc 
+      volumes: ➏
+         name: mypvc
```
fabiand: a dash is missing, ti sahould be:

```
- name: mypvc
```

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+         presistentVolumeClaim:
+      claimName: rhel-7-pvc-${NAME}
+- apiVersion: kubevirt.io/v1alpha1
+  kind: presistentVolumeClaim ❼
+  metadata:
+    name: rhel-7-pvc-${NAME} ❽
+labels: ...       
+parameters: ...
```
fabiand: Here you should name CPU, MEMORY, and NAME

docs/virtual_machine_template_guide.adoc
fabiand: @rmohr do you have kubevirt docs describing that we support quantities?

IOW Is there a PR describing the VM API design approach?

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+
+❹ OfflineVirtualMachine’s resources, currently includes only memory and cpu. +
+   Cpu is the number of cpu cores the Virtual Machine requires. +
+   Memory is the minimal amount the Virtual Machine requires. +
+   These attributes are represented as parameters that will either be filled out by the user Or receive default values. +
+   
+➎ Disk contains a "volumeName" attribute, which links it to a volume +
+
+➏ Specification of a disk’s volume from the disks section. +
+        It contains a persistent volume claim.  (more about link:https://kubernetes.io/docs/concepts/storage/persistent-volumes/[Persistent Volume Claims] ) +
```
fabiand: Maybe an incorrect indentration?

docs/virtual_machine_template_guide.adoc
fabiand: Take a look @fromanirh 's temaplte examples.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,195 @@
...
+        volumeName: mypvc 
+      volumes: ➏
+         name: mypvc
+         presistentVolumeClaim:
+      claimName: rhel-7-pvc-${NAME}
+- apiVersion: kubevirt.io/v1alpha1
+  kind: presistentVolumeClaim ❼
+  metadata:
+    name: rhel-7-pvc-${NAME} ❽
+labels: ...
```
fabiand: labels should be part of the template metadata very for up

docs/virtual_machine_template_guide.adoc
fabiand: /cc @fromanirh

docs/virtual_machine_template_guide.adoc
yanirq: its elaborated under parameters

docs/virtual_machine_template_guide.adoc
rmohr: "now" it is there.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,197 @@
...
+kind: Template
+metadata:
+  name: RHEL-7 ❶
+  labels:
+    miq.github.io/kubvirt/is-vm-template: True
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine ❷
+  metadata:
+    name: rhel-7-${NAME} ❸
```
fabiand: Should be equal to NAME

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,197 @@
...
+              name: disk0
+              volumeName: mypvc 
+      volumes: ❼
+       - name: mypvc
+         presistentVolumeClaim:
+           claimName: rhel-7-pvc-${NAME}
+- apiVersion: kubevirt.io/v1alpha1
+  kind: PersistentVolumeClaim ❽
+  metadata:
+    name: rhel-7-pvc-${NAME} ➒
```
fabiand: here we need an annotation to say which volume this pvc should be based on

@aglitke do you have an example how a PVC should look to request a clone?

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,197 @@
...
+
+
+=== Full VirtualMachine Template example
+
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha1
+kind: Template 
+metadata:
+  name: RHEL-7  #This will define the template’s name
```
fabiand: The comment can be omitted

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,197 @@
...
+  # name: Microsoft-Windows-2012r1
+  labels:
+    miq.github.io/kubvirt/is-vm-template: true  
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine
+  metadata:
+    name: rhel-7-${NAME}
```
fabiand: name should be equal to NAME

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,197 @@
...
+              name: disk0
+              volumeName: mypvc 
+      volumes:
+       - name: mypvc
+         presistentVolumeClaim:
+           claimName: rhel-7-pvc-${NAME}
+- apiVersion: kubevirt.io/v1alpha1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: rhel-7-pvc-${NAME}
```
fabiand: Here we also need to include the annotation to request a clone @aglitke - got an example?

fabiand: The previous is a non blocking item for merging this - but we should then add a FIXME note

docs/virtual_machine_template_guide.adoc
fabiand: The previous is a non blocking item for merging this - but we should then add a FIXME note

yanirq: @fabiand  added FIXME,  an issue can also be opened for that matter if needed

fabiand: Awesome, thank you!

pkliczewski:
@yanirq Please squash your commits. It would be great if you could update the [example](https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/manifests/example-template.yml).

pkliczewski:
Actually the example could be updated in different PR so please squash.

yanirq:
@pkliczewski  commits are squashed

pkliczewski:
@rmohr Are you still requiring changes to this PR?

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha1
+kind: Template
+metadata:
+  name: RHEL-7 ❶
+  annotations:
+  labels:
+    miq.github.io/kubvirt-os: rhel-7 ❷
```
masayag: Shouldn't it be ```kubevirt-os``` and ```kubevirt-is-vm-template``` ?

fabiand:
lgtm

docs/virtual_machine_template_guide.adoc
yanirq: done

pkliczewski:
@rmohr I see that you set require changes flag. Is it still valid?

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,209 @@
...
+The following parameters will be used for the OfflineVirtualMachine name (and as a part of the persistent volume claim names) +
+Default values will be provided for memory and cpu, which will be used if the user does not supply a different value.
+
+[source,yaml]
+----
+parameters:
+- description: Name for the new VM     
+  name: NAME   
+- name: MEMORY
+  description: Amount of memory in MB
```
rmohr: This description is wrong

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+If you have a YAML file that defines a VirtualMachine template, you can upload the template to an Openshift cluster using the CLI. 
+To upload a template to your current project’s template library, pass the YAML file with the following command: +
+
+----
+$ oc create -f <filename>
+----
+
+
+=== Creating VirtualMachineTemplate Objects using the Web Console
```
rmohr: VirtualMachineTemplate does not exists as a k8s kind. it is just an open shift template.

yanirq: @rmohr 
so if its an issue with the overall phrasing  (OfflineVirtualMachine and VirtualMachineTemplate)
changing OfflineVirtualMachine to "virtual machine template" (without mentioning its an ocp template - it is already mentioned at the beginning)  will be the right change ? 

also, are you also suggesting removing OfflineVirtualMachine from "kind:" in code examples ?

rmohr: OfflineVirtualMachine is fine since it exists (in the text and in "kind"). Using "virtual machine template" instead of VirtualMachineTemplate solves my issue.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,209 @@
...
+            cores: ${CPU}
+          devices:
+            disks: ❽
+              name: disk0
+              volumeName: mypvc 
+      volumes: ➒
+       - name: mypvc
+         presistentVolumeClaim:
+           claimName: rhel-7-pvc-${NAME}
+- apiVersion: kubevirt.io/v1alpha1
```
pkliczewski: PVC is not part of kubevirt api version. We should use `apiVersion: v1`

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,209 @@
...
+  kind: PersistentVolumeClaim ➓
+  metadata:
+    name: rhel-7-pvc-${NAME} ⓫
+    #FIXME: add an annotation to specify which volume this pvc should be based on
```
pkliczewski: This is not enough information to create pvc. Kubernetes is complaining about:

> * spec.accessModes: Required value: at least 1 access mode is required
> * spec.resources[storage]: Required value

Adding following spec solved the issue for me:
```yaml
   spec:
    accessModes:
    - ReadWriteOnce
    resources:
      requests:
        storage: 1Gi
```

fabiand: @yanirq can you include this?

yanirq: @pkliczewski  @fabiand  done

pkliczewski:
@yanirq We still have one more person requiring changes.

docs/virtual_machine_template_guide.adoc
```diff
@@ -0,0 +1,222 @@
...
+        
+➓ The PersistenVolumeClaim Object that accompanies the OfflineVirtualMachine Object. +
+
+⓫ PersistenVolumeClaims should include: name, permissions and requested amount of storage. +
+
+
+
+=== Labels
+
+Labels will be presented on every OCP VirtualMachineTemplate in order for ManageIQ to recognize the KubeVirt template characteristics. +
```
rmohr: VirtualmachineTemplate still exists in the text.

yanirq: @rmohr  done

PR #9, opened 2018-01-10
docs: add template examples for kubevirt

fabiand:
Add example template to build OfflineVirtualMachine
for linux and windows flavours.

### DIFF
--- a/docs/template-vm.yaml
+++ b/docs/template-vm.yaml
@@ -13,12 +13,15 @@ objects:
   spec:
     template:
       metadata:
-      labels:
+        labels:
+          miq.github.io/kubevirt-is-vm-template: true
       spec:
         domain:
           resources:
-            memory: ${MEMORY}
-            cpu: ${CPU}
+            requests:
+              memory: ${MEMORY}
+              cpu:
+                cores: ${CPU}
           os:
             machine: pc-q35-rhel7.4.0
           devices:
@@ -31,16 +34,16 @@ objects:
                 disk:
                   dev: vda
                   bus: virtio
-  volumes:
-    - name: disk0-pvc
-      persistentVolumeClaim:
-        claimName: linux-vm-pvc-${NAME}
+        volumes:
+          - name: disk0-pvc
+            persistentVolumeClaim:
+              claimName: linux-vm-pvc-${NAME}
 - apiVersion: kubevirt.io/v1alpha1
   kind: PersistentVolumeClaim
   metadata:
     name: linux-vm-pvc-${NAME}
-labels:
-  miq.github.io/kubevirt/is-vm-template: True
+ labels:
+  miq.github.io/kubevirt-is-vm-template: True
 parameters:
 - description: Name for the new VM
   name: NAME
--- a/docs/template-vm.yaml
+++ b/docs/template-vm.yaml
@@ -1,0 +1,52 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: linux-vm-template
+  annotations:
+    description: "OCP kubevirt linux, template"
+    tags: "kubevirt,ocp,template,linux"
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine
+  metadata:
+    name: linux-vm-${NAME}
+  spec:
+    template:
+      metadata:
+      labels:
+      spec:
+        domain:
+          resources:
+            memory: ${MEMORY}
+            cpu: ${CPU}
+          os:
+            machine: pc-q35-rhel7.4.0
+          devices:
+            interface:
+              model:
+                type: virtio
+            disks:
+              - name: disk0
+                volumeName: disk0-pvc
+                disk:
+                  dev: vda
+                  bus: virtio
+  volumes:
+    - name: disk0-pvc
+      persistentVolumeClaim:
+        claimName: linux-vm-pvc-${NAME}
+- apiVersion: kubevirt.io/v1alpha1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: linux-vm-pvc-${NAME}
+labels:
+  miq.github.io/kubevirt/is-vm-template: True
+parameters:
+- description: Name for the new VM
+  name: NAME
+- name: MEMORY
+  description: Amount of memory in MB
+  value: 4096Mi
+- name: CPU
+  description: Amount of cores
+  value: 4
--- a/docs/template-vm.yaml
+++ b/docs/template-vm.yaml
@@ -19,7 +19,7 @@ objects:
             requests:
               memory: ${MEMORY}
               cpu:
-                cores: ${CPU}
+                cores: ${CPU_CORES}
           os:
             machine: pc-q35-rhel7.4.0
           devices:
@@ -46,6 +46,6 @@ parameters:
 - name: MEMORY
   description: Amount of memory in MB
   value: 4096Mi
-- name: CPU
+- name: CPU_CORES
   description: Amount of cores
   value: 4
--- a/docs/template-vm.yaml
+++ b/docs/template-vm.yaml
@@ -20,8 +20,8 @@ objects:
           resources:
             requests:
               memory: ${MEMORY}
-          os:
-            machine: pc-q35-rhel7.4.0
+          machine:
+            type: q35
           devices:
             interface:
               model:
--- a/docs/template-vm.yaml
+++ b/docs/template-vm.yaml
@@ -5,6 +5,8 @@ metadata:
   annotations:
     description: "OCP kubevirt linux, template"
     tags: "kubevirt,ocp,template,linux"
+  labels:
+    miq.github.io/kubevirt-is-vm-template: true
 objects:
 - apiVersion: kubevirt.io/v1alpha1
   kind: OfflineVirtualMachine
@@ -44,5 +46,3 @@ parameters:
 - name: CPU_CORES
   description: Amount of cores
   value: 4
-labels:
-  miq.github.io/kubevirt-is-vm-template: true
--- a/docs/template-vm.yaml
+++ b/docs/template-vm.yaml
@@ -15,11 +15,11 @@ objects:
     template:
       spec:
         domain:
+          cpu:
+            cores: ${CPU_CORES}
           resources:
             requests:
               memory: ${MEMORY}
-              cpu:
-                cores: ${CPU_CORES}
           os:
             machine: pc-q35-rhel7.4.0
           devices:
--- a/docs/template-vm.yaml
+++ b/docs/template-vm.yaml
@@ -5,7 +5,6 @@ metadata:
   annotations:
     description: "OCP kubevirt linux, template"
     tags: "kubevirt,ocp,template,linux"
-    miq.github.io/kubevirt-is-vm-template: true
 objects:
 - apiVersion: kubevirt.io/v1alpha1
   kind: OfflineVirtualMachine
@@ -23,15 +22,11 @@ objects:
           machine:
             type: q35
           devices:
-            interface:
-              model:
-                type: virtio
             disks:
               - name: disk0
                 volumeName: disk0-pvc
                 disk:
                   dev: vda
-                  bus: virtio
         volumes:
           - name: disk0-pvc
             persistentVolumeClaim:
@@ -49,3 +44,5 @@ parameters:
 - name: CPU_CORES
   description: Amount of cores
   value: 4
+labels:
+  miq.github.io/kubevirt-is-vm-template: true
--- a/docs/template-vm.yaml
+++ b/docs/template-vm.yaml
@@ -5,16 +5,14 @@ metadata:
   annotations:
     description: "OCP kubevirt linux, template"
     tags: "kubevirt,ocp,template,linux"
+    miq.github.io/kubevirt-is-vm-template: true
 objects:
 - apiVersion: kubevirt.io/v1alpha1
   kind: OfflineVirtualMachine
   metadata:
-    name: linux-vm-${NAME}
+    name: ${NAME}
   spec:
     template:
-      metadata:
-        labels:
-          miq.github.io/kubevirt-is-vm-template: true
       spec:
         domain:
           resources:
@@ -42,8 +40,6 @@ objects:
   kind: PersistentVolumeClaim
   metadata:
     name: linux-vm-pvc-${NAME}
- labels:
-  miq.github.io/kubevirt-is-vm-template: True
 parameters:
 - description: Name for the new VM
   name: NAME
--- a/docs/template-windows.yaml
+++ b/docs/template-windows.yaml
@@ -5,7 +5,6 @@ metadata:
   annotations:
     description: "OCP kubevirt windows VM template"
     tags: "kubevirt,ocp,template,windows"
-    miq.github.io/kubevirt-is-vm-template: true
 objects:
 - apiVersion: kubevirt.io/v1alpha1
   kind: OfflineVirtualMachine
@@ -41,15 +40,11 @@ objects:
             requests:
               memory: ${MEMORY}
           devices:
-            interface:
-              model:
-                type: e1000
             disks:
               - name: disk0
                 volumeName: disk0-pvc
                 disk:
                   dev: sda
-                  bus: sata
         volumes:
           - name: disk0-pvc
             persistentVolumeClaim:
@@ -67,3 +62,5 @@ parameters:
 - name: CPU_CORES
   description: Amount of cores
   value: 4
+labels:
+  miq.github.io/kubevirt-is-vm-template: true
--- a/docs/template-windows.yaml
+++ b/docs/template-windows.yaml
@@ -5,16 +5,14 @@ metadata:
   annotations:
     description: "OCP kubevirt windows VM template"
     tags: "kubevirt,ocp,template,windows"
+    miq.github.io/kubevirt-is-vm-template: true
 objects:
 - apiVersion: kubevirt.io/v1alpha1
   kind: OfflineVirtualMachine
   metadata:
-    name: windows-vm-${NAME}
+    name: ${NAME}
   spec:
     template:
-      metadata:
-        labels:
-          miq.github.io/kubevirt-is-vm-template: true
       spec:
         domain:
           features:
@@ -62,8 +60,6 @@ objects:
   kind: PersistentVolumeClaim
   metadata:
     name: windows-vm-pvc-${NAME}
-labels:
-  miq.github.io/kubevirt-is-vm-template: True
 parameters:
 - description: Name for the new VM
   name: NAME
--- a/docs/template-windows.yaml
+++ b/docs/template-windows.yaml
@@ -13,9 +13,9 @@ objects:
   spec:
     template:
       metadata:
-      labels:
+        labels:
+          miq.github.io/kubevirt-is-vm-template: true
       spec:
-        terminationGracePeriodSeconds: 0
         domain:
           features:
             acpi: {}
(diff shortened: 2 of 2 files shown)


fabiand:
Created this pr in order to be able to add comments.

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: linux-vm-template
+  annotations:
+    description: "OCP kubevirt linux, template"
+    tags: "kubevirt,ocp,template,linux"
```
fabiand: Here the annotation is missing to tell MIQ that this is a tempalte @yanirq can you provide the right annotation?

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
...
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine
+  metadata:
+    name: linux-vm-${NAME}
+  spec:
+    template:
+      metadata:
+      labels:
```
fabiand: This needs to be indentet - we can actually drop metadata including labels all together as it's empty.

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
...
+      spec:
+        domain:
+          resources:
+            memory: ${MEMORY}
+            cpu: ${CPU}
```
fabiand: Here you should probably use the CPU API as introduced here https://github.com/kubevirt/kubevirt/pull/638

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
...
+          os:
+            machine: pc-q35-rhel7.4.0
```
fabiand: Hopefulyl we will default to this (or similar) so that it can be omitted @mpolednik

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
...
+            interface:
+              model:
+                type: virtio
+            disks:
+              - name: disk0
+                volumeName: disk0-pvc
+                disk:
+                  dev: vda
+                  bus: virtio
+  volumes:
```
fabiand: volumes are part of the spec, thusn eeds to be indented further

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - name: disk0-pvc
+      persistentVolumeClaim:
+        claimName: linux-vm-pvc-${NAME}
+- apiVersion: kubevirt.io/v1alpha1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: linux-vm-pvc-${NAME}
+labels:
```
fabiand: this label would be on the template metadat further above (where I commented that it is missing)

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: linux-vm-template
```
fabiand: Can we use a specific name, like rhel-7-template

Then it will be clearer further below when we reference disk images.

docs/template-windows.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: windows-vm-template
```
fabiand: I'd also make this a little clearer windows-2012r2-vm-template I guess.

docs/template-windows.yaml
```diff
@@ -0,0 +1,73 @@
...
+  annotations:
+    description: "OCP kubevirt windows VM template"
+    tags: "kubevirt,ocp,template,windows"
```
fabiand: Here the tempalte annotation is also needed.

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
...
+  annotations:
+    description: "OCP kubevirt linux, template"
+    tags: "kubevirt,ocp,template,linux"
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine
+  metadata:
+    name: linux-vm-${NAME}
```
fabiand: I actually wonder how the name should be.

As a user I'd input "greenhorn" as the NAME - Then, the vm linux-vm-greenhorn will be created. Makes sense. but do we want this behavior, or that a vm called "greenhorn" is getting created?

docs/template-windows.yaml
```diff
@@ -0,0 +1,73 @@
...
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine
+  metadata:
+    name: windows-vm-${NAME}
```
fabiand: Same reasoning about the name as above - should itname be == to NAME variable or with this prefix?

docs/template-windows.yaml
```diff
@@ -0,0 +1,73 @@
...
+  spec:
+    template:
+      metadata:
+      labels:
```
fabiand: labels need to be indentet a little more, but as emtpy we can remove metadat alltogether

docs/template-windows.yaml
```diff
@@ -0,0 +1,73 @@
...
+            timer:
+              hpet:
+                present: false
+              pit:
+                tickPolicy: delay
+              rtc:
+                tickPolicy: catchup
+              hyperv: {}
+          firmware:
+            uid: 5d307ca9-b3ef-428c-8861-06e72d69f223
```
fabiand: Tricky. This would lead to all windows vms to have the same firwmare uuid.

I wonder if we could have a variable as well here. Maybe CFME has a uuid field? @jhernand

docs/template-windows.yaml
```diff
@@ -0,0 +1,73 @@
...
+            interface:
+              model:
+                type: e1000
+            disks:
+              - name: disk0
+                volumeName: disk0-pvc
+                disk:
+                  dev: sda
+                  bus: sata
+      volumes:
```
fabiand: As above, needs the same indentation level as domain

docs/template-windows.yaml
```diff
@@ -0,0 +1,73 @@
...
+        - name: disk0-pvc
+          persistentVolumeClaim:
+            claimName: windows-vm-pvc-${NAME}
+- apiVersion: kubevirt.io/v1alpha1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: windows-vm-pvc-${NAME}
+labels:
```
fabiand: Needs to go into the Template metadata

docs/template-vm.yaml
yanirq: @fabiand @fromanirh 
see https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/8
I used a label for that matter: 

labels:
  //vm-template label applied to all objects created from this template
  template: “miq.github.io/kubvirt/vm-template”

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
...
+  spec:
+    template:
+      metadata:
```
petrkotas: In the template, do we count that the OfflineVirtualMachine name will be derived from template automatically? If not, the name should be defined here.

docs/template-vm.yaml
petrkotas: What is the role of labels here? Is the OfflineVirtualMachine controller supposed to generate them automatically?

docs/template-vm.yaml
fromanirh: Ok, I placed in the wrong spot, will move above

docs/template-vm.yaml
fromanirh: I though about naming it rhel-7, but no settings was rhel-7 specific, so I choose to be more generic. I can switch to rhel-7 if reviewers prefer.

docs/template-vm.yaml
fromanirh: sure, the annotation is above, will move here

docs/template-vm.yaml
fromanirh: good question, I'm open to all alternatives :)

docs/template-vm.yaml
fromanirh: I think this relates to what @fabiand commented right above

docs/template-vm.yaml
fromanirh: Will remove all the groups as suggested

docs/template-vm.yaml
fromanirh: Will do

docs/template-vm.yaml
fromanirh: I'm not sure about the defaults so I choose to be explicit

docs/template-windows.yaml
fromanirh: or maybe windows-2016? I'm fine with this choice, but AFAIK nothing is windows-2012 specific (meaning, it should work with every windows starting from 2012)

docs/template-windows.yaml
fromanirh: right, good catch

docs/template-vm.yaml
yanirq: on one hand, it will make the name,pvcs and maybe other attributes in the future more consistent (a consistent suffix)
on the other hand we do limit the user here with a predefined prefix (unless he edits the yaml completely) 
The former is what we have in the user guide for now

docs/template-vm.yaml
petrkotas: Currently OfflineVirtualMachine does not require the labels. That is the reason of my curiosity :)

docs/template-vm.yaml
yanirq: @fabiand @fromanirh   
The labels are at the bottom of the file on purpose, and in the form of:

labels:  
      **template**: “miq.github.io/kubvirt/vm-template”  

“miq.github.io/kubvirt/vm-template”  label applied to all objects created from this template 
and without having a flag of true/flase

the syntax is mentioned here:
https://docs.openshift.org/latest/dev_guide/templates.html#writing-labels

I'm open to other suggestions/corrections here

docs/template-windows.yaml
yanirq: see comment above

docs/template-vm.yaml
yanirq: @petrkotas  well it was picked up from the Offline virtual machine doc :)  can be removed if not needed

docs/template-vm.yaml
yanirq: add also requests under resources

docs/template-vm.yaml
fromanirh: according to the discussion in https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/8, I added the ^metadata.label$ above. Also, according to https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/8 looks like we need to use a different format, with only one "slash".

docs/template-vm.yaml
fromanirh: For the time being still kept linux-vm-template, will do another upload switching to rhel-7 if needed later

docs/template-vm.yaml
fromanirh: done

docs/template-vm.yaml
fromanirh: done

docs/template-windows.yaml
fromanirh: open to any suggestion, not changed for the moment

docs/template-vm.yaml
fromanirh: BTW, I was following openshift guidelines here: https://docs.openshift.org/latest/dev_guide/templates.html#writing-labels
not sure how those labels (not the metadata.labels) with enter into play.

docs/template-vm.yaml
jhernand: I think that the name of the `OfflineVirtualMachine` created from the template should be exactly what the user provides in the `NAME` variable:

```yaml
name: ${NAME}
```

Otherwise we will have to explain to users that the names of virtual machines are somehow calculated from the name of the template and the value that they provider for `NAME`. It can be quite confusing for users.

docs/template-vm.yaml
```diff
@@ -0,0 +1,52 @@
...
+            disks:
+              - name: disk0
+                volumeName: disk0-pvc
+                disk:
+                  dev: vda
+                  bus: virtio
+  volumes:
+    - name: disk0-pvc
+      persistentVolumeClaim:
+        claimName: linux-vm-pvc-${NAME}
```
jhernand: Unlike the name of the `OfflineVirtualMachine`, the name of the claim, and the name of the persistent volume, can be calculated, with a prefix if you find it appropriate, because the user won't care about those names, actually we won't show them in the UI (I mean in the ManageIQ UI).

docs/template-windows.yaml
jhernand: In my opinion it should be exactly equal to `NAME`.

docs/template-windows.yaml
jhernand: ManageIQ has an internal identifier for the virtual machine. But it isn't an UUID, and it is assigned *after* the `OfflineVirtualMachine` is created: ManageIQ receives the notification from Kubernetes and then creates its own VM object.

If really required, then ManageIQ could generate an UUID when creating the `OfflineVirtualMachine`, but that is the kind of detail that the Kubevirt client should not be aware of. Can't we just remove this attribute from the template? Will Kubevirt assign then a new UUID to the `OffileVirtualMachine`?

docs/template-vm.yaml
fabiand: wfm

docs/template-windows.yaml
fabiand: Ugh - tehse windows version. klet's keep it as is :)

docs/template-windows.yaml
fabiand: same comment as above

docs/template-windows.yaml
fabiand: Yes, we can do that.

We can remove it from the template, and have an initializer who is adding it to the OfflineVM upon creation if not specified.

@rmohr does this sound ok?

fabiand: /cc @petrkotas

docs/template-vm.yaml
fromanirh: no problem, changing accordingly (both templates)

docs/template-windows.yaml
fromanirh: done

docs/template-vm.yaml
fabiand: As discussed above - The NAME parameter should be used as the name for the OfflineVM

docs/template-vm.yaml
fabiand: WFM

docs/template-vm.yaml
```diff
@@ -0,0 +1,51 @@
...
+  metadata:
+    name: ${NAME}
+  spec:
+    template:
+      spec:
+        domain:
+          resources:
+            requests:
+              memory: ${MEMORY}
+              cpu:
```
rmohr: that needs to go under `spec.domain.cpu.cores`. `resources.requests.cpu` is about cgroup assignment

fromanirh: Ack, thanks for catching this.

docs/template-windows.yaml
```diff
@@ -0,0 +1,71 @@
...
+            timer:
+              hpet:
+                present: false
+              pit:
+                tickPolicy: delay
+              rtc:
+                tickPolicy: catchup
+              hyperv: {}
+          firmware:
+            uid: ${UID}
```
rmohr: UID does not seem to be reflected in the parameters. 

Just for completeness: If the uid is not specified, OfflineVirtualMachine can/will try to provide an UID based on the VM name. That has its limits (e.g. vms in differnet namespaces with the same name can get the same UID, which in theory could produce licensing  issues) but it is something which we might for instance use in a future VirtualMachineStatefulSet.

fabiand: +1

Let's dro the uuid from this teomplate for now, we'll sovle it in the scope of OfflineVM

fromanirh: Done

docs/template-windows.yaml
```diff
@@ -0,0 +1,69 @@
...
+                tickPolicy: delay
+              rtc:
+                tickPolicy: catchup
+              hyperv: {}
+          cpu:
+            cores: ${CPU_CORES}
+          resources:
+            requests:
+              memory: ${MEMORY}
+          os:
```
rmohr: According to https://github.com/kubevirt/kubevirt/pull/650, it is `machine.type` and not `os.machine`.

fromanirh: Fixed. I'm not sure about the supported values. "q35" surely is, and it is good enough, so I'm using it.

docs/template-vm.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: linux-vm-template
+  annotations:
+    description: "OCP kubevirt linux, template"
+    tags: "kubevirt,ocp,template,linux"
+    miq.github.io/kubevirt-is-vm-template: true
```
rmohr: That should be a label, so that manageiq can directly query for supported templates.

docs/template-vm.yaml
```diff
@@ -0,0 +1,51 @@
...
+        domain:
+          cpu:
+            cores: ${CPU_CORES}
+          resources:
+            requests:
+              memory: ${MEMORY}
+          machine:
+            type: q35
+          devices:
+            interface:
```
rmohr: We do not yet allow setting any interfaces. Right we add one by default which  is not visible to the user. That will for sure change in the furute, but it will no have effect now.

docs/template-vm.yaml
```diff
@@ -0,0 +1,51 @@
...
+              model:
+                type: virtio
+            disks:
+              - name: disk0
+                volumeName: disk0-pvc
+                disk:
+                  dev: vda
+                  bus: virtio
```
rmohr: Such a `bus` field does not yet exist.

docs/template-vm.yaml
fromanirh: OK, will remove it

docs/template-vm.yaml
fromanirh: not needed, will remove it

docs/template-vm.yaml
fromanirh: ok, moving as label

docs/template-windows.yaml
```diff
@@ -0,0 +1,66 @@
...
+parameters:
+- description: Name for the new VM
+  name: NAME
+- name: MEMORY
+  description: Amount of memory in MB
+  value: 4096Mi
+- name: CPU_CORES
+  description: Amount of cores
+  value: 4
+labels:
```
rmohr: `labels` needs to be part of the template metadata: `metadata.labels`.

docs/template-vm.yaml
pkliczewski: Is parameter describing OS details?

docs/template-vm.yaml
```diff
@@ -0,0 +1,48 @@
...
+  labels:
+    miq.github.io/kubevirt-is-vm-template: true
+objects:
+- apiVersion: kubevirt.io/v1alpha1
+  kind: OfflineVirtualMachine
+  metadata:
+    name: ${NAME}
+  spec:
+    template:
+      spec:
```
pkliczewski: Do we want to allow networks here?

fabiand: No, networks are out of scope for now.

docs/template-vm.yaml
fabiand: No. The field (name) is just defining the template name - it is technically not tied to functionality, just an identifier.

pkliczewski: so how are we going to identify OS version?

fabiand: Some time ago we had the discussion which metadata was needed. Name, CPU, memory, flag for identifing applicable templates were the ones named.

It sounds as if you also need to understand which OS is used by a template, is this correct? Do you need a way to extract this information?

fromanirh:
@pkliczewski : done in another PR to avoid increasing the mess here : https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/17 . Do you prefer me to go ahead in pull/17 or should I update this one (and abandon the new)?

pkliczewski:
@fromanirh I am with going with #17

PR #10, opened 2018-01-14
Vm start and stop

pkliczewski:
Few fixes to enable vm power management operations such as stop and start.
Fixed by @jhernand.


pkliczewski:
@jhernand @masayag please review

app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
```diff
@@ -82,4 +82,12 @@ def generate_remote_viewer_file
     # Return the generated file:
     file
   end
+
+  #
+  # UI Button Validation Methods
+  #
+
+  def has_required_host?
```
jhernand: Maybe we could explain here why this is needed, so that the next person reading this code isn't as puzzled by this as we were.

PR #11, opened 2018-01-16
Adjust the memory calculator to Kubevirt 0.2.0

masayag:
In Kubevirt 0.2.0 the way that virtual machine memory is specified has
changed. In previous versions it was specified like this:

  memory:
    unit: MB
    value: 64

See the example here:

  https://github.com/kubevirt/kubevirt/blob/v0.1.0/cluster/vm.yaml#L34-L36

Since version 0.2.0 it is specified like this:

  memory: 64M

See the example here:

  https://github.com/kubevirt/kubevirt/blob/v0.2.0/cluster/vm.yaml#L10

This means that the `MemoryCalculator` class needs to be changed
accordingly to support the new format.

Also the provision request and the parsing of the reported virtual machines were updated to support the updated memory format.

Signed-off-by: Juan Hernandez <juan.hernandez@redhat.com>
Signed-off-by: Moti Asayag <masayag@redhat.com>

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
@@ -65,10 +65,7 @@ module ManageIQ::Providers::Kubevirt::InfraManager::Provision::Cloning
           template: {
             spec: {
               domain: {
-                memory: {
-                  value: memory.to_i,
-                  unit: 'MiB'
-                }
+                memory: memory.to_s + 'Mi'
               }
             }
           }
--- a/app/models/manageiq/providers/kubevirt/inventory/parser.rb
+++ b/app/models/manageiq/providers/kubevirt/inventory/parser.rb
@@ -181,7 +181,7 @@ class ManageIQ::Providers::Kubevirt::Inventory::Parser < ManagerRefresh::Invento
 
     # Create the inventory object for the hardware:
     hw_object = hw_collection.find_or_build(vm_object)
-    hw_object.memory_mb = ManageIQ::Providers::Kubevirt::MemoryCalculator::convert(domain.memory.value, domain.memory.unit, 'MiB')
+    hw_object.memory_mb = ManageIQ::Providers::Kubevirt::MemoryCalculator::convert(domain.memory, 'Mi')
 
     # Return the created inventory object:
     vm_object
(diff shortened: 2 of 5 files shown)


masayag:
@pkliczewski please review

app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
```diff
@@ -65,10 +65,7 @@ def start_clone(options)
           template: {
             spec: {
               domain: {
-                memory: {
-                  value: memory.to_i,
-                  unit: 'MiB'
-                }
+                memory: memory.to_s + 'MiB'
```
jhernand: According to the Kubenetes documentation the recommended way to specify the memory in MiB is using the `Mi` suffix, without the `B`. See here for details: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory

app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -181,7 +181,7 @@ def process_domain(domain, uid, name)
 
     # Create the inventory object for the hardware:
     hw_object = hw_collection.find_or_build(vm_object)
-    hw_object.memory_mb = ManageIQ::Providers::Kubevirt::MemoryCalculator::convert(domain.memory.value, domain.memory.unit, 'MiB')
+    hw_object.memory_mb = ManageIQ::Providers::Kubevirt::MemoryCalculator::convert(domain.memory, 'MiB')
```
jhernand: Same, consider using just `Mi`.

app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -213,7 +213,7 @@ def process_template(object)
 
     # Add the inventory object for the hardware:
     hw_object = hw_collection.find_or_build(template_object)
-    hw_object.memory_mb = ManageIQ::Providers::Kubevirt::MemoryCalculator::convert(domain.memory.value, domain.memory.unit, 'MiB')
+    hw_object.memory_mb = ManageIQ::Providers::Kubevirt::MemoryCalculator::convert(domain.memory, 'MiB')
```
jhernand: Same, just `Mi`.

manifests/example-template.yml
```diff
@@ -33,9 +33,7 @@ spec:
             type: network
           video:
           - type: qxl
-        memory:
-          unit: MB
-          value: 64
+        memory: 64M
```
jhernand: Note that this is 64\*10^6 bytes. You may want to use 64MiB which is 64\*2^30 bytes.

masayag:
@jhernand I've applied your comments for this PR. Thanks for the review.

PR #14, opened 2018-01-18
Convert badges to .adoc format

chessbyte:


pkliczewski:
We do not need to convert badges to .adoc format. We decided to be consistent with other repos and converted README to markdown. Closing this PR since we merged https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/15

PR #18, opened 2018-01-24
Don't put `RestClient` in debug mode

jhernand:
Currently the provider puts the `RestClient` class in debug mode, doing
the following:

```ruby
RestClient.log = 'stdout'
```

This is a leftover of the initial development of the provider, when it
was very convenient for debugging. But now it should be removed, because
it affects this provider and also any other providers that use the same
gem.


jhernand:
@pkliczewski @masayag please review.

PR #19, opened 2018-01-25
Rubocop auto-fix

masayag:


pkliczewski:
it looks like we have 2 offences. Please fix and we can merge.

masayag:
@pkliczewski there is nothing we can do with those offenses. 

The first is defined on the core, and the second was designed this way.
These issues weren't introduced by this PR, the lines were changed therefore - the CI get to complain for those.

There are additional issues that I'll handle manually in a follow-up patch.

PR #21, opened 2018-01-25
Use `:resource_version` instead of `resourceVersion`

jhernand:
The parameter that the `kubeclient` gem uses to specify the resource
version for a watch is the `:resource_version` symbol, not the
`resourceVersion` string:

  https://github.com/abonas/kubeclient/blob/v2.4.0/lib/kubeclient/common.rb#L246

But the provider is using the string `resourceVersion` instead, which is
silently ignored.


jhernand:
@pkliczewski @masayag please review.

PR #22, opened 2018-01-25
fix rubocop complaints

masayag:


app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -111,6 +111,8 @@ def process_node(object)
       :host    => host_object,
       :storage => storage_object,
     )
+
+    host_storage_object
```
jhernand: I think this isn't correct. Why should the `process_node` method return the `host_storage_object`? If Rubocop complains about that I'd suggest to just return `nil` explicitly.

masayag: what I should have done is just removing the assignment to `host_storage_object`

app/models/manageiq/providers/kubevirt/memory_calculator.rb
```diff
@@ -47,8 +47,6 @@ def self.convert(value, to)
     value * from_factor / to_factor
   end
 
-  private
-
```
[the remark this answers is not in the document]
masayag: since it's  a static method, it should be declared as `private_class_method :scale_factor` to become private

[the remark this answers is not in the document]
jhernand: That makes sense. What is important is to make sure that these constants and method aren't used outside of this class.

app/models/manageiq/providers/kubevirt/inventory/parser.rb
jhernand: Yes. You may still want to return `nil` explicitly, because otherwise the method will return the same object anyhow. Not really important, as nothing uses the value returned by this method.

masayag:
@jhernand I've addressed your comments. please review.

PR #24, opened 2018-01-29
Support retire VM

masayag:
Kubevirt supports retiring a Vm, regardless its state by removing its
live vm (for running vm) and its offline vm.


pkliczewski:
I would love to see a spec for this code

masayag:
@pkliczewski updated PR with specs.

PR #25, opened 2018-01-29
Add VM Provision dialog

masayag:
Kubevirt is a virtualization provider that allows to create virtual
machines. Currently, it is pretty limited and not fully defined,
therefore the initial dialog allows the user to specific the VM memory
only.

For vm provisioning, the :linked_clone attribute is ignored, as the
decision of disk provision type is made by the storage backend.
In addition, we don't control the node on which the vm will be
scheduled, therefore the implementation of allowed_hosts is irrelevant in this
context.

The core PR that adds kubevirt's template to the eligible templates list:
https://github.com/ManageIQ/manageiq/pull/16873

**Request Tab**
vm_provision_request_tab

**Catalog Tab**
vm_provision_catalog_tab

**Hardware Tab**
vm_provision_hardware_tab

**Schedule Tab**
vm_provision_schedule_tab

**Invalid name error message**
vm_provision_invalid_name

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/provision_workflow.rb
@@ -35,10 +35,6 @@ class ManageIQ::Providers::Kubevirt::InfraManager::ProvisionWorkflow < MiqProvis
     get_value(@values[:provision_type]).to_s == 'native_clone'
   end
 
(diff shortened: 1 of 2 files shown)


pkliczewski:
@masayag the change looks good, can you please fix codeclimate issue?

pkliczewski:
Let's wait for core PR with merging this one

PR #28, opened 2018-02-07
Refresh removed vm

pkliczewski:
During retirement process we need to target removed vm during refresh.
This PR fixes the issue with refresh opration.

### DIFF
--- a/app/models/manageiq/providers/kubevirt/builder.rb
+++ b/app/models/manageiq/providers/kubevirt/builder.rb
@@ -22,13 +22,44 @@
 class ManageIQ::Providers::Kubevirt::Builder
   class << self
     def build_inventory(manager, target)
+      persister = ManageIQ::Providers::Kubevirt::Inventory::Persister.new(manager, target)
+
       # Create the collector, parser and persister:
-      collector = ManageIQ::Providers::Kubevirt::Inventory::Collector.new(manager, target)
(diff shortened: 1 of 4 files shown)


pkliczewski:
@masayag @jhernand please review

app/models/manageiq/providers/kubevirt/builder.rb
```diff
@@ -22,10 +22,32 @@
...
+        collector.nodes = {}
+        manager.with_provider_connection do |connection|
+          collector.offline_vms = connection.offline_vm(name)
+          collector.live_vms = connection.live_vm(name)
+          collector.templates = connection.template(name)
+        end
+
+        parser = ManageIQ::Providers::Kubevirt::Inventory::Parser::PartialRefresh.new
+      else
+        collector = ManageIQ::Providers::Kubevirt::Inventory::Collector.new(manager, manager)
```
masayag: this looks very similar to https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/app/models/manageiq/providers/kubevirt/infra_manager/refresh_worker/runner.rb#L118

Can we reuse this builder for creating the full refresh collector, parser and persister ?
This will allow us also to cover the same code with the exact spec file.

pkliczewski: Good idea, will do.

PR #30, opened 2018-02-22
Make templates consumable by openshift

pkliczewski:
There were issues with our example templates. This patch fixes them so
we can use `oc` tool to create them.


pkliczewski:
@fromanirh @fabiand please review

fromanirh:
LGTM - and I didn't know that everything must be a "string".

PR #40, opened 2018-03-05
Watch restart when expiration notice received

pkliczewski:
We can receive following notices from Kubernetes/Openshift:

#<Kubeclient::Common::WatchNotice type="ERROR",
 object={:kind=>"Status",
         :apiVersion=>"v1",
         :metadata=>{},
         :status=>"Failure",
         :message=>"too old resource version: 1203002 (1691766)",
         :reason=>"Gone",
         :code=>410}>

#<Kubeclient::Common::WatchNotice type="ERROR",
 object={:kind=>"Status",
         :apiVersion=>"v1",
         :metadata=>{},
         :status=>"Failure",
         :message=>"The resourceVersion for the provided watch is too old.",
         :reason=>"Expired",
         :code=>410}>

For both we need to restart watches since we do not know which object's version
is old. Similar situation occurs when MIQ was down for longer period of time
and we need to run every time full_refresh on MIQ start.


pkliczewski:
@masayag @jhernand Please review

PR #41, opened 2018-03-06
Use guest os label

pkliczewski:
There is a change in label name how we expose os information.

This PR was pushed because https://github.com/kubevirt/user-guide/pull/25 was merged.

### DIFF
--- a/app/models/manageiq/providers/kubevirt/inventory/parser.rb
+++ b/app/models/manageiq/providers/kubevirt/inventory/parser.rb
@@ -33,6 +33,11 @@ class ManageIQ::Providers::Kubevirt::Inventory::Parser < ManagerRefresh::Invento
   #
   STORAGE_ID = '0'.freeze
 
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -236,7 +236,7 @@ def process_hardware(template_object, params, metadata, domain)
     cpu = default_value(params, 'CPU_CORES')
     hw_object.cpu_cores_per_socket = cpu
     hw_object.cpu_total_cores = cpu
-    hw_object.guest_os = metadata.labels.send("miq.github.io/kubevirt-os")
+    hw_object.guest_os = metadata.labels.send("kubevirt.io/os")
```
masayag: please extract `kubevirt.io/os` to a constant an reuse it in line 272 as well.

PR #42, opened 2018-03-06
Role kubevirt-infra was removed

pkliczewski:
We do not need kubevirt-infra role in our manifests since it was removed.
It is enough to use cluster-admin.

### DIFF
--- a/manifests/account-kubernetes.yml
+++ b/manifests/account-kubernetes.yml
@@ -17,16 +17,3 @@ subjects:
   name: miq-account
   namespace: default
 ---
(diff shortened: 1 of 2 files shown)


masayag:
@pkliczewski can't we remove those file completely ?

pkliczewski:
@masayag we seems to use them from time to time. We could replace them with kubectl or oc command line in the user guide but it needs to be done.

masayag:
@pkliczewski in that case, let's merge this one and add task to replace it with user guide.

Issue #44, opened 2018-03-07
404 error with KubeVirt virtualization


vatsalparekh:
When adding the OpenShift provider, and selecting the KubeVirt virtualization, if the openshift doesn't have kubevirt but the token for Virtualization validation is successful, we get the error
`Credential validation was not successful: HTTP status code 404, 404 Not Found`

vatsalparekh:
@masayag

masayag:
The reason for the 404 is due to the use of `kubevirt_client` in order to access the `versions`.
Since kubevirt isn't deployed on the server, the action fails.

The fix should replace `kubevirt_client` with `core_client` in:
`api_versions = kubevirt_client.api["versions"]`

https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/app/models/manageiq/providers/kubevirt/infra_manager/connection.rb#L94

PR #46, opened 2018-03-07
Use core client to query for supported versions

masayag:
In order to avoid 404 when kubevirt isn't deployed, the core client
should be used to check if kubevirt version is available.

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/44


pkliczewski:
@masayag Looks good, let me know when it is verified.

masayag:
@pkliczewski please review.

Issue #48, opened 2018-03-08
Targeted refresh for VM created via Miq shows error


vatsalparekh:
Doing targeted refresh for VMs created via Miq (lifecycle) on KubeVirt results in error in Last refresh pane
```

Error - Less Than A Minute Ago
--
HTTP status code 404, templates "vparekh-test-linux" not found

```

PR #52, opened 2018-03-19
Target refresh

masayag:
Implement target refresh for vms, offline vms and templates.

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/48

### DIFF
--- a/app/models/manageiq/providers/kubevirt/builder.rb
+++ b/app/models/manageiq/providers/kubevirt/builder.rb
@@ -30,12 +30,20 @@ class ManageIQ::Providers::Kubevirt::Builder
         collector = ManageIQ::Providers::Kubevirt::Inventory::Collector.new(manager, target)
         collector.nodes = {}
         manager.with_provider_connection do |connection|
-          collector.offline_vms = connection.offline_vm(name)
-          collector.live_vms = connection.live_vm(name)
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/kubevirt/builder.rb
```diff
@@ -30,12 +30,19 @@ def build_inventory(manager, target)
...
-          collector.live_vms = connection.live_vm(name)
-          collector.templates = connection.template(name)
+          if target.template?
+            collector.templates = [connection.template(name)]
+          else
+            collector.offline_vms = [connection.offline_vm(name)]
+            begin
+              collector.live_vms = [connection.live_vm(name)]
+            rescue
+              # target refresh of an offline vm might fail if it has no live vm
```
pkliczewski: should we log here something here. Maybe in debug.

pkliczewski:
@masayag what do you think about small spec to cover it?

app/models/manageiq/providers/kubevirt/builder.rb
masayag: added log

masayag:
> @masayag what do you think about small spec to cover it?

I"ll add spec for it

PR #62, opened 2018-04-20
Use fog-kubevirt

pkliczewski:


app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -277,32 +253,32 @@ def process_hardware(template_object, params, metadata, domain)
...
+      if param[:name] == name
+        value = param[:value]
       end
     end
     value
   end
 
   def process_disks(hw_object, domain)
-    domain.devices.disks.each do |disk|
+    domain.dig(:devices, :disks).each do |disk|
```
borod108: I think this might through an error if  domain.dig(:devices, :disks) is nil

app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -277,32 +253,32 @@ def process_hardware(template_object, params, metadata, domain)
   def default_value(params, name)
     value = nil
     params.each do |param|
-      if param.name == name
-        value = param.value
+      if param[:name] == name
+        value = param[:value]
```
borod108: maybe this method can be something like:

name_param = params.detect { |param| param[:name] == name }
named_param[:value]

although to tell the truth I do not understand exactly what this is supposed to do, since it seems like it might be overwriting value several times.

app/models/manageiq/providers/kubevirt/inventory/parser/partial_refresh.rb
```diff
@@ -76,16 +69,12 @@ def parse
   private
 
   def get_object_ids(notices)
-    ids = notices.map { |notice| notice.object.metadata.uid }
+    ids = notices.map { |notice| notice.metadata.uid }
```
borod108: I think we can replace these three lines with one:
notices.map { |notice| notice.metadata.uid }.uniq

app/models/manageiq/providers/kubevirt/inventory/parser/partial_target_refresh.rb
```diff
@@ -60,7 +60,7 @@ def parse
   private
 
   def get_object_ids(objects)
```
borod108: same as above
objects.map { |o| o.uid }.uniq

spec/models/manageiq/providers/kubevirt/infra_manager/provision/cloning_spec.rb
```diff
@@ -21,46 +21,24 @@
...
-    it "creates specific objects with pvc" do
+    it "calls clone on template" do
       source = FactoryGirl.create(:template_kubevirt)
       subject.source = source
       connection = double("connection")
 
-      allow(connection).to receive(:template).and_return(unprocessed_object("template.json"))
+      template = double("template")
+      allow(template).to receive(:name).and_return("example")
+      expect(template).to receive(:clone)
```
borod108: if this expectation is supposed to be fulfilled by the last line of the test, I would move it to the line before last, I could not figure out if there is another expect missing in this spec.

spec/models/manageiq/providers/kubevirt/inventory/parse_vm_spec.rb
```diff
@@ -46,7 +46,16 @@
       parser.instance_variable_set(:@hw_collection, hw_collection)
       parser.instance_variable_set(:@network_collection, network_collection)
 
-      parser.send(:process_live_vm, unprocessed_object("vm.json"))
+      source = double("VirtualMachine")
```
borod108: maybe just:
source = double(:uid => "asdf", :name => "asdf".....)

spec/models/manageiq/providers/kubevirt/inventory/parser_spec.rb
```diff
@@ -45,7 +45,18 @@
       parser.instance_variable_set(:@vm_os_collection, os_collection)
       parser.instance_variable_set(:@disk_collection, disk_collection)
 
-      parser.send(:process_template, unprocessed_object("template.json"))
+      # TODO: check whether this is the format that fog-kubevirt would return
+      json = unprocessed_object("template.json")
+
+      source = double("template")
```
borod108: same as above

spec/models/manageiq/providers/kubevirt/inventory/parser_spec.rb
```diff
@@ -104,7 +115,18 @@
       parser.instance_variable_set(:@vm_os_collection, os_collection)
       parser.instance_variable_set(:@disk_collection, disk_collection)
 
-      parser.send(:process_template, unprocessed_object("template_registry.json"))
+      # TODO: check whether this is the format that fog-kubevirt would return
+      json = unprocessed_object("template_registry.json")
+
+      source = double("template")
```
borod108: same

spec/models/manageiq/providers/kubevirt/inventory/parser_spec.rb
```diff
@@ -45,7 +45,18 @@
       parser.instance_variable_set(:@vm_os_collection, os_collection)
       parser.instance_variable_set(:@disk_collection, disk_collection)
 
-      parser.send(:process_template, unprocessed_object("template.json"))
+      # TODO: check whether this is the format that fog-kubevirt would return
```
borod108: is this an important #TODO? if so maybe it should be done now, it looks like a simple puts can check it. Else maybe it is not important and then we can drop it.

app/models/manageiq/providers/kubevirt/inventory/parser.rb
pkliczewski: will add a check

app/models/manageiq/providers/kubevirt/inventory/parser.rb
pkliczewski: will check whether you solution would work

app/models/manageiq/providers/kubevirt/inventory/parser/partial_refresh.rb
pkliczewski: Done

app/models/manageiq/providers/kubevirt/inventory/parser/partial_target_refresh.rb
pkliczewski: Done

spec/models/manageiq/providers/kubevirt/infra_manager/provision/cloning_spec.rb
pkliczewski: OK

spec/models/manageiq/providers/kubevirt/inventory/parse_vm_spec.rb
pkliczewski: Done

spec/models/manageiq/providers/kubevirt/inventory/parser_spec.rb
pkliczewski: Done

spec/models/manageiq/providers/kubevirt/inventory/parser_spec.rb
pkliczewski: Done

spec/models/manageiq/providers/kubevirt/inventory/parser_spec.rb
pkliczewski: Old comment, will remove

app/models/manageiq/providers/kubevirt/inventory/parser.rb
pkliczewski: I do not see how it can be `nil` so there is no need to have additional check here.

pkliczewski:
@agrare Do you see any issues in this PR dependency of `fog-kubevirt`?

agrare:
@pkliczewski nope LGTM

Issue #74, opened 2018-05-16
VM not getting powered on


vatsalparekh:
>Note: This happens time to time

Provisioning request of a VM gets stuck in 'Active', checking it on openshift, the offline VM was created, but it was not powered on, didn't find anything openshift logs and events.
in manageiq, 
`automation.log`
```
[----] I, [2018-05-16T04:56:12.453707 #2157:124d140]  INFO -- : Q-task_id([r9_miq_provision_request_9]) Processed State=[limits]
[----] W, [2018-05-16T04:56:12.453789 #2157:124d140]  WARN -- : Q-task_id([r9_miq_provision_request_9]) Skipping to state finished
```
But the request state in UI was still `Active`

vatsalparekh:
This was checked on appliance with #62

pkliczewski:
I checked the flow and noticed that initial refresh fails due to 

> [----] E, [2018-05-17T14:08:03.100348 #27290:4ba9ad8] ERROR -- : [RuntimeError]: Please provide Hash as a reference, :manager_ref count includes more than 1 attribute. keys: [:hardware, :device_name], data: c8524a39-52c2-11e8-b1b2-525400a36119 Method:[block in method_missing]

This potentially could be related to: https://github.com/ManageIQ/manageiq/commit/fc2b8e81e2dc81200612b3606fa3cfc31b389e06

When I removed one key from [here](https://github.com/ManageIQ/manageiq/blob/e915a3f7ef305c4a3d12cdb06b37e1d39902f6b2/app/models/manager_refresh/inventory_collection_default.rb#L172) the refresh works but vm state is not reflected correctly.

@Ladas Can you please take a look?

pkliczewski:
The issue was [here](https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/app/models/manageiq/providers/kubevirt/inventory/parser/partial_refresh.rb#L63). [This](https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/68/files#diff-56a97fc380d5562a1c3cd8a8e831fee4) was not complete.

pkliczewski:
[...]

PR #78, opened 2018-05-21
Finalizing target refresh cleanup

pkliczewski:
There was a PR merged which not fully cleanup disk collection code used
during target refresh.


pkliczewski:
I split this out of https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/62 and it fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/74

Issue #77, opened 2018-05-21
validating Virtualization token gives 401 error


vatsalparekh:
When editing the container provider, if the default was invalid, and it was validated in this change, validating virtualization with correct token would give 401 error

vatsalparekh:
seems to be working now!

PR #81, opened 2018-05-23
Respect kubevirt token if provided

masayag:
When the token for kubevirt is being provided (i.e. when changing the token by the user), it should be respected and used to verify connectivity and the virtualization support instead of relying on the existing token.

Required by https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/256
Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/77


cben:
can you clarify "provided is being provided" in PR description :)

masayag:
> can you clarify "provided is being provided" in PR description :)

@cben updated the PR description.

Issue #69, opened 2018-05-14
Gap in Power state of a VM


vatsalparekh:
Unlike the traditional providers, Kubevirt has VM scheduling like Kubernetes way, means VMs would have phases like `Scheduling`, `Scheduled`,  `Pending`, or `Failed` in some cases.
The gap here is, even if the VM scheduling is failed, but it stays in `vms`, so in miq we see it as `Powered On` but that's not powered on actually. Or even `Scheduling` is like booting up for traditional vm, but here we just see it as `Powered on`, which is actually not the actual state.

vatsalparekh:
this seems like current gap for miq, but maybe can be a gap for all the integrations using KubeVirt as a VM provider.
@pkliczewski @fabiand your inputs?

pkliczewski:
@vatsalparekh Very good point. There are two potential solutions for this issue:

- Map kubevirt states to those defined in MIQ
- Add kubevirt specific states to MIQ

@agrare With which one we should go?

agrare:
@pkliczewski I think we should have the real values in the raw_power_state column but map to traditional VM power states for consistency with the rest of the infra/cloud providers as well as we already have icons to represent these states.

pkliczewski:
@agrare Thank you! Will use the approach you suggested.

vatsalparekh:
@pkliczewski Seeing one more gap that sometimes the power state of the VM differ the actual state on the provider, but working on this issue would fix it

pkliczewski:
PR solving it was merged

PR #82, opened 2018-05-25
Use vm status provided by kubevirt

pkliczewski:
We assumed about vm status based on part of the flow we were in.
This PR uses vm status as collected from kubevirt.

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
@@ -17,8 +17,18 @@
 class ManageIQ::Providers::Kubevirt::InfraManager::Vm < ManageIQ::Providers::InfraManager::Vm
   include_concern 'Operations'
 
+  POWER_STATES = {
+    'Running'    => 'on',
+    'Pending'    => 'powering_up',
+    'Scheduling' => 'powering_up',
+    'Scheduled'  => 'off',
+    'Succeeded'  => 'off',
+    'Failed'     => 'off',
+    'Unknown'    => 'terminated'
+  }.freeze
+
   def self.calculate_power_state(raw)
-    raw
+    POWER_STATES[raw] || super
   end
 
   #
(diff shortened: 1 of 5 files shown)


pkliczewski:
This PR fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/69

pkliczewski:
This PR requires https://github.com/fog/fog-kubevirt/pull/12 to work.

app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
```diff
@@ -17,8 +17,17 @@
 class ManageIQ::Providers::Kubevirt::InfraManager::Vm < ManageIQ::Providers::InfraManager::Vm
   include_concern 'Operations'
 
+  POWER_STATES = {
+    'Running'    => 'on',
+    'Pending'    => 'powering_up',
+    'Scheduling' => 'powering_up',
```
pkliczewski: @rmohr Please check whether I added here all the statuses that are possible for VMs. I based my list on pod life cycle https://blog.openshift.com/kubernetes-pods-life/

rmohr: Right now we have these: https://github.com/kubevirt/kubevirt/blob/master/pkg/api/v1/types.go#L303. The only one you are missing is `Scheduled`. That one does not exist on pods.

I also saw now that these fields are not reflected in the api-reference: http://www.kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachinestatus. Will file an issue in kubevirt.

rmohr: And we should probably rethink our `Scheduled` and `Pending` and `Scheduling` meaning. We slightly diverge here from Pods, with out necessarily for a good reason. But for now it works like that.

rmohr: For reference: https://github.com/kubevirt/kubevirt/issues/1085

pkliczewski: Great, I will update it with `Scheduled`. Feel free to open an issue if anything changes with vm statuses.

Issue #38, opened 2018-03-05
Infra provider not getting updated


vatsalparekh:
After adding the `Openshift` provider with `kubevirt` for virtualization, I was able to see the infra provider, but that provider does not get refreshed like the container provider does. I've to manually do the refresh for the provider to fetch the data.
![screenshot from 2018-03-05 16-02-40](https://user-images.githubusercontent.com/6758282/36971050-8cca6be8-2090-11e8-8fc6-28e29e393e0f.png)

pkliczewski:
Can you please specify the steps to reproduce?

vatsalparekh:
With the solved issue #39 , I am seeing the VM and template info getting updated in the infra provider, but the field `Last Refresh` stays `Never`.

Steps:
- Add KubeVirt with Openshift as Container provider
- Check the Infra provider of Kubevirt

jhernand:
@pkliczewski note that this is a side effect of the refresh architecture of the provider. It doesn't use the traditional refresh mechanism used by other ManageIQ providers, so it never updates that "Last Refresh" field.

pkliczewski:
@jhernand Thanks. Looks like a gap we need to fix at some point. Needs to be investigated whether it is possible with our current code base.

pkliczewski:
PR fixing the issue was merged

PR #83, opened 2018-05-29
refresh: update last refresh time

pkliczewski:
We want to see when last refresh was triggered.


pkliczewski:
@agrare please review

pkliczewski:
This fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/38

Issue #71, opened 2018-05-14
Save button not getting enabled when Editing Infra provider


vatsalparekh:
When editing virtualization endpoint for a token change, the save button is not getting enabled, even after waiting for quite few minutes
![screenshot from 2018-05-14 18-03-34](https://user-images.githubusercontent.com/6758282/39997660-4b3f1f78-57a1-11e8-9824-7a22fc35bd52.png)

vatsalparekh:
@pkliczewski this is specifically for #62

vatsalparekh:
@pkliczewski Because of this, I can't check #72 for #62 
Would suggest if you can fix this one before merging #62

pkliczewski:
It looks like it works from time to time. We need to understand what is breaking it.

pkliczewski:
I checked and when there is not correct token in parent provider we are unable to save valid token in the child provider.

vatsalparekh:
From the latest findings, the reproducible pattern is that, if the auth status for `default` token of openshift is still valid, then save button will appear when editing the infra provider, but if the status is `!` then it will not let you save the new correct token for virtualization.
Checked it without the fog-kubevirt, so this is not regression, but a bug rather.

masayag:
This issue will be fixed by https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/256

vatsalparekh:
@masayag @pkliczewski this one is still reproducible on latest master

masayag:
Reopen the issue - should be closed once https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/259 is merged.

masayag:
The only use case in which the issue might occur after merging the previous PRs is when the parent provider token is invalid, while kubevirt's token is valid.

The assumption is that both kubevirt and the container manager provider tokens should be valid. 
The options are to verify parent provider token is valid when checking kubevirt validation or to disable the ability to edit kubevirt provider.

The other use-cases are handled by the merged PRs.

PR #93, opened 2018-06-13
Implement auth methods

masayag:
The default_auth_status is being used by the UI to detect the validity
of the provider. For kubevirt, the default auth_status refers to the
parent provider, while kubevirt auth_status is detected by the :kubevirt
auth_type.

For that purpose, there is a need to report always 'true' for the
'default_auth_status', so the kubevirt provider can be edited when being
accessed as an infra provider.

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/71


masayag:
@pkliczewski I split this PR into https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/96 until https://github.com/ManageIQ/manageiq/pull/17576#issuecomment-398317864 is resolved.

Since kubernetes PR is merged, we can merge this one as well.

PR #94, opened 2018-06-18
Rename VM entities names

masayag:


app/models/manageiq/providers/kubevirt/infra_manager/vm/operations.rb
```diff
@@ -21,16 +21,16 @@ module ManageIQ::Providers::Kubevirt::InfraManager::Vm::Operations
 
   def raw_destroy
     ext_management_system.with_provider_connection do |connection|
-      # Retrieve the details of the offline virtual machine:
+      # Retrieve the details of the virtual machine:
       begin
-        live_vm = connection.live_vm(name)
+        vm_instance = connection.vm_instance(name)
       rescue KubeException
```
pkliczewski: Maybe not in this PR but we need to remove KubeException here. I wonder whether it would be worth to run kubeclient exception in fog-kubevirt specific exception.

masayag: @pkliczewski we actually have a card to track that task.

masayag: @pkliczewski I'd take this one step further and eliminate the dependency of miq-providers-kubevirt on kubeclient. I think that we shouldn't have a direct dependency on kubeclient at all.

pkliczewski: :+1:

masayag:
@pkliczewski seems to fetch the latest fog-kubevirt-0.1.3.gem correctly

PR #97, opened 2018-06-19
Remove unused method

masayag:


masayag:
@pkliczewski  Without https://github.com/ManageIQ/manageiq/pull/17576#issuecomment-398317864 there is no need in default_auth_status_ok? method. We can remove it at the meantime and if accepted, to implement it.

PR #98, opened 2018-06-20
Remove useless EmsRefresherMixin

agrare:
Remove useless EmsRefresherMixin removed in https://github.com/ManageIQ/manageiq/pull/17472


agrare:
@pkliczewski can you take a look, this is the last provider that we need to remove this empty include from before we can delete the module from core

agrare:
@miq-bot assign @pkliczewski

agrare:
Thanks @masayag !

PR #112, opened 2018-08-14
Add an explicit unknown location

masayag:
Since storage wasn't introduced to fog-kubevirt, there is no information
to report on the location of vms and templates.
However, since the 'location' field is mandatory, preserving the former
'unknown' value explicitly.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1610772


masayag:
@pkliczewski note I haven't added a spec file since this is being covered by:

https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/spec/models/manageiq/providers/kubevirt/inventory/parse_vm_spec.rb#L69

PR #149, opened 2019-10-02
Add methods for for provider creation pluggable UI

agrare:
Add a method indicating the properties required to create a Kubevirt
provider and implement the common self.verify_credentials interface
method.

https://github.com/ManageIQ/manageiq/issues/18818

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager.rb
@@ -81,7 +81,7 @@ class ManageIQ::Providers::Kubevirt::InfraManager < ManageIQ::Providers::InfraMa
             {
               :type             => "validatorTypes.MAX_NUMBER_VALUE",
               :includeThreshold => true,
-              :value            => 65535
+              :value            => 65_535
             }
           ]
         },
@@ -98,7 +98,7 @@ class ManageIQ::Providers::Kubevirt::InfraManager < ManageIQ::Providers::InfraMa
   end
 
   def self.verify_credentials(args)
-    !!raw_connect(args.dig("endpoints", "default")&.values_at("server", "port", "token")&.symbolize_keys)
+    !!raw_connect(args.dig("endpoints", "default")&.slice("server", "port", "token")&.symbolize_keys)
   end
 
   #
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -53,6 +53,41 @@ def self.description
...
+          :label      => "Hostname",
+          :isRequired => true,
+          :validate   => [{:type => "required-validator"}]
+        },
+        {
+          :component  => "text-field",
+          :name       => "endpoints.default.port",
+          :type       => "number",
+          :isRequired => true,
+          :validate   => [{:type => "required-validator"}]
```
Fryguy: I think there's a validator for numbers.

agrare: You mean for a port range like `[1, 65535]`?  Yeah can add that

app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -53,6 +53,51 @@ def self.description
...
+        },
+        {
+          :component  => "text-field",
+          :name       => "endpoints.default.port",
+          :type       => "number",
+          :isRequired => true,
+          :validate   => [
+            {:type => "required-validator"},
+            {:type             => "validatorTypes.MIN_NUMBER_VALUE",
```
agrare: Not sure if this is right, @Hyperkid123 ?

Hyperkid123: Yes that seems correct.

You can see live example with same use case just different values here if you want to see it live: https://data-driven-forms.org/renderer/validators#numbervaluevalidators

agrare: It lines up with https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/149/files#diff-ff17b05a7c525eec833f0ddb0134a0d9R75

app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -53,6 +53,51 @@ def self.description
...
+          :isRequired => true,
+          :validate   => [{:type => "required-validator"}]
+        }
+      ]
+    }.freeze
+  end
+
+  def self.verify_credentials(args)
+    server, port, token = args.dig("endpoints", "default")&.values_at("server", "port", "token")
+    !!raw_connect(:server => server, :port => port, :token => token)
```
Fryguy: Super minor but if the incoming hash already has the keys...this can be simplified to just use slice.

```ruby
!!raw_connect(args.dig("endpoints", "default")&.slice("server", "port", "token").symbolize_keys)
```

Fryguy: Also, interestingly, the docs for raw_connect say it takes verify_ssl and ca_certs, but then never uses them? https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/149/files#diff-ff17b05a7c525eec833f0ddb0134a0d9R109-R111

agrare: Yeah I couldn't tell if that was an oversight or not so I left the description for now.

PR #150, opened 2019-10-23
Subclassed EmsCluster to ManageIQ::Providers::Kubevirt::InfraManager:Cluster

chessbyte:


agrare:
We need to update the collection definition [here](https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/app/models/manageiq/providers/kubevirt/inventory/persister.rb#L22), [here](https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/spec/models/manageiq/providers/kubevirt/inventory/full_refresh_spec.rb#L108), and [here](https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/spec/models/manageiq/providers/kubevirt/inventory/partial_target_refresh_spec.rb#L113)

app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -79,7 +79,7 @@ def process_node(object)
     # Add the inventory object for the host:
     host_object = host_collection.find_or_build(uid)
     host_object.connection_state = 'connected'
-    host_object.ems_cluster = cluster_collection.lazy_find(CLUSTER_ID)
+    host_object.cluster = cluster_collection.lazy_find(CLUSTER_ID)
```
agrare: Hm I don't think this will work, the association is still [`belongs_to :ems_cluster`](https://github.com/ManageIQ/manageiq/blob/master/app/models/host.rb#L32)

agrare: Yeah,
```
  1) ManageIQ::Providers::Kubevirt::Inventory::Parser::FullRefresh#parse works correctly with one node
     Failure/Error: host_object.cluster = cluster_collection.lazy_find(CLUSTER_ID)
     
     NoMethodError:
       undefined method `cluster=' for #<#<Class:0x00005594b207cf78>:0x00005594b22c6a18>
     # ./app/models/manageiq/providers/kubevirt/inventory/parser.rb:82:in `process_node'
```

spec/models/manageiq/providers/kubevirt/inventory/full_refresh_spec.rb
```diff
@@ -54,7 +54,7 @@
       expect(host.vmm_product).to eq('KubeVirt')
       expect(host.vmm_vendor).to eq('kubevirt')
       expect(host.vmm_version).to eq('0.1.0')
-      expect(host.ems_cluster).to_not be_nil
+      expect(host.cluster).to_not be_nil
```
agrare: Same, this should be host.ems_cluster

agrare:
This one still needs to be resolved, https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/spec/models/manageiq/providers/kubevirt/inventory/full_refresh_spec.rb#L108

After that one and those two comments are resolved this repo is green with your other changes

chessbyte:
@agrare I think I addressed all your concerns.  Please double-check

PR #157, opened 2019-11-14
Remove ems_ref_obj

agrare:
Setting ems_ref_obj = ems_ref is no longer necessary


agrare:
@gtanzillo can you take a look?

PR #161, opened 2020-04-21
BigDecimal#initialize was removed in 2.0.0

agrare:
Move to BigDecimal(value) instead of BigDecimal.new(value) as this was removed and no longer works on recent rubies (https://github.com/ruby/bigdecimal/blob/master/CHANGES.md#200)

https://github.com/ManageIQ/manageiq/issues/19678


agrare:
Thanks for the review @masayag this is green now

PR #192, opened 2021-11-24
convert Ems#supports_create to supports :create

kbrock:
part of https://github.com/ManageIQ/manageiq/pull/21573


kbrock:
this defaults as not able to create
no reason to `supports_not`

PR #195, opened 2022-01-31
Fix KubeVirt endpoint verification

agrare:
`connection.valid?` doesn't return the same connection which causes a later check for `#virt_supported?` to return `nil` when the connection actually does support virtualization.

The result was even with valid endpoint info and credentials the credential verification would always fail with "unknown error"

```
(byebug) kubevirt
#<ManageIQ::Providers::Kubevirt::InfraManager::Connection:0x0000559626830880 @namespace="default", ...
(byebug) kubevirt.valid?
#<Kubeclient::Resource kind="Namespace", apiVersion="v1", ...
(byebug) kubevirt.valid?.virt_supported? # This is essentially what https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/457/files#diff-1fe0d067134e4f6092c7cf36d13501a6d283dcc890e7279c0de4059af6a49de5L817 was doing
nil
```

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager.rb
@@ -114,7 +114,8 @@ class ManageIQ::Providers::Kubevirt::InfraManager < ManageIQ::Providers::InfraMa
   end
 
   def self.verify_credentials(args)
-    !!raw_connect(args.dig("endpoints", "default")&.slice("server", "port", "token")&.symbolize_keys)
+    kubevirt = raw_connect(args.dig("endpoints", "default")&.slice("server", "port", "token")&.symbolize_keys)
+    kubevirt&.valid? && kubevirt&.virt_supported?
   end
 
   #
@@ -128,14 +129,11 @@ class ManageIQ::Providers::Kubevirt::InfraManager < ManageIQ::Providers::InfraMa
   #
   def self.raw_connect(opts)
     # Create the connection:
-    connection = Connection.new(
+    Connection.new(
       :host  => opts[:server],
       :port  => opts[:port],
       :token => ManageIQ::Password.try_decrypt(opts[:token])
     )
-
-    # Verify that the connection works:
-    connection.valid?
   end
 
   #


app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -128,14 +129,11 @@ def self.verify_credentials(args)
...
     # Create the connection:
-    connection = Connection.new(
+    Connection.new(
       :host  => opts[:server],
       :port  => opts[:port],
       :token => ManageIQ::Password.try_decrypt(opts[:token])
     )
-
-    # Verify that the connection works:
-    connection.valid?
```
agrare, line 138: This was doing some of the validation but ended up not returning the `connection` object.  Rewritten to have raw_connect simply return the connection and moved all verification logic into `.verify_credentials` above

app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -114,7 +114,8 @@ def self.params_for_create
   end
 
   def self.verify_credentials(args)
-    !!raw_connect(args.dig("endpoints", "default")&.slice("server", "port", "token")&.symbolize_keys)
+    kubevirt = raw_connect(args.dig("endpoints", "default")&.slice("server", "port", "token")&.symbolize_keys)
+    kubevirt&.valid? && kubevirt&.virt_supported?
```
agrare, line 118: The verification logic was kind of split between the kubernetes and kubevirt repos, kubevirt was doing `#valid?` and kubernetes was doing `#virt_supported?` [[ref]](https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/457/files#diff-1fe0d067134e4f6092c7cf36d13501a6d283dcc890e7279c0de4059af6a49de5L817)

Seemed much simpler to just move all verify logic into this method

Fryguy:
Backported to `morphy` in commit 2b3f08a5782e22cf3edb6ed9044d9a441b8a12aa.

```text
commit 2b3f08a5782e22cf3edb6ed9044d9a441b8a12aa
Author: Keenan Brock <keenan@thebrocks.net>
Date:   Mon Jan 31 12:43:43 2022 -0500

    Merge pull request #195 from agrare/fix_kubevirt_endpoint_verification
    
    Fix KubeVirt endpoint verification
    
    (cherry picked from commit 664c55ab22e0104a9407d9d51476cf8c302a34a2)
```

PR #203, opened 2022-03-01
Handle float iec60027_2 values like "1.5Gi"

agrare:
We have a more_core_extensions `iec_60027_2_to_i` method but it also doesn't handle float, need to do more research to see if this is something we should fix in that extension or if this is kubevirt specific but I do have a kubevirt template with `1.5Gi` of memory


Fryguy:
> need to do more research to see if this is something we should fix in that extension

More than likely we'd need a separate method for floats, like `iec_60027_2_to_f` (detail being: to_f vs to_i)

Fryguy:
Is it possible to add a spec for this?

agrare:
> More than likely we'd need a separate method for floats, like iec_60027_2_to_f (detail being: to_f vs to_i)

Well it isn't really to a float, this is still a decimal number after it is expanded out to bytes

Fryguy:
Oh I see what you mean... yeah, then we'd have to deal with it in this branch of the logic: https://github.com/ManageIQ/more_core_extensions/blob/master/lib/more_core_extensions/core_ext/string/iec60027_2.rb#L13

Fryguy:
Released 4.4.0 of more_core_extensions that includes https://github.com/ManageIQ/more_core_extensions/pull/106

app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -164,14 +164,24 @@ def process_domain(memory, cores, uid, name)
...
     # Return the created inventory object:
     vm_object
   end
 
+  def parse_quantity(value)
+    return nil if value.nil?
+
+    begin
+      value.iec_60027_2_to_i
+    rescue
```
Fryguy, line 180: Should we just rescue ArgumentError only here?  Not sure.

agrare: So I pulled this from the kubernetes parser, figure we can refactor to pull that out into a module in the future and this is the way it handles the rescue

app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -164,14 +164,24 @@ def process_domain(memory, cores, uid, name)
 
     # Create the inventory object for the hardware:
     hw_object = hw_collection.find_or_build(vm_object)
-    hw_object.memory_mb = ManageIQ::Providers::Kubevirt::MemoryCalculator.convert(memory, 'Mi')
+    hw_object.memory_mb = parse_quantity(memory) / 1.megabytes.to_f
```
Fryguy, line 167: Do we have tests exercising this for different extensions?  Only asking because I tried some stuff locally and it fails with expected values as per the original tests.

```
irb(main):008:0> "1 Ki".iec_60027_2_to_i
=> 1024
irb(main):009:0> "1 KB".iec_60027_2_to_i
Traceback (most recent call last):
        4: from bin/console:14:in `<main>'
        3: from (irb):9
        2: from /Users/jfrey/dev/more_core_extensions/lib/more_core_extensions/core_ext/string/iec60027_2.rb:11:in `iec_60027_2_to_i'
        1: from /Users/jfrey/dev/more_core_extensions/lib/more_core_extensions/core_ext/string/iec60027_2.rb:11:in `Integer'
ArgumentError (invalid value for Integer(): "1 KB")
irb(main):010:0> "1 KB".decimal_si_to_f
Traceback (most recent call last):
        5: from bin/console:14:in `<main>'
        4: from (irb):9
        3: from (irb):10:in `rescue in irb_binding'
        2: from /Users/jfrey/dev/more_core_extensions/lib/more_core_extensions/core_ext/string/decimal_suffix.rb:10:in `decimal_si_to_f'
        1: from /Users/jfrey/dev/more_core_extensions/lib/more_core_extensions/core_ext/string/decimal_suffix.rb:10:in `Float'
ArgumentError (invalid value for Float(): "1 KB")
```

I expected the last one to "work", but maybe I'm passing the wrong values?

agrare: I don't think that "1 KB" is the format that kubernetes uses, here is how they parse capacity fields https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/7ba567b0d2af94c71a85ccb54c4dde088f128e99/app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb#L817-L836

Fryguy: Interesting.  Wonder why the specs here had strings like `"1KB"`, then.

agrare: yeah I have no idea, if you want I can put back my original fix which just updates the regex

app/models/manageiq/providers/kubevirt/inventory/parser.rb
Fryguy: +1 for moving this out into a shared module, then this provider can reuse it from the kube provider.

Fryguy:
Backported to `najdorf` in commit 8ac219fcf7052a6c73b880254cd6812d9cd0ddff.

```text
commit 8ac219fcf7052a6c73b880254cd6812d9cd0ddff
Author: Jason Frey <fryguy9@gmail.com>
Date:   Wed Mar 2 11:46:58 2022 -0500

    Merge pull request #203 from agrare/fix_memory_calculator_float
    
    Handle float iec60027_2 values like "1.5Gi"
    
    (cherry picked from commit 0214a51bfa650356d940b79b6594747f5692ba2e)
```

PR #210, opened 2022-08-10
Don't rely on rails magic to find fixtures, be explicit

jrafanie:
As of https://github.com/ManageIQ/manageiq/pull/22039
core now specifies the fixture path so we need to be explicit when running
these tests.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -17,3 +17,7 @@ VCR.configure do |config|
     config.define_cassette_placeholder(secrets.kubevirt_defaults[secret]) { secrets.kubevirt[secret] }
   end
 end
+
+RSpec.configure do |config|
+  config.file_fixture_path = ManageIQ::Providers::Kubevirt::Engine.root.join("spec/fixtures/files")
+end


Fryguy:
Is this something we should put in all of the providers?  Perhaps at the provider generator level?  Or perhaps in core in a pluggable manner?  cc @agrare

Fryguy:
Removed najdorf/yes? as we are not backporting anything Rails 6.1

agrare:
Yeah I see a number of providers have `spec/fixtures`:
```
manageiq-providers-amazon/spec/fixtures:
orchestration_templates

manageiq-providers-azure/spec/fixtures:
events  orchestration_templates

manageiq-providers-azure_stack/spec/fixtures:
orchestration_templates

manageiq-providers-kubevirt/spec/fixtures:
files

manageiq-providers-openstack/spec/fixtures:
orchestration_templates

manageiq-providers-vmware/spec/fixtures:
orchestration_templates
```

I'm assuming all of these would have to be updated?

As for the generator, if `spec/fixtures` aren't used we probably don't need this...or should we create an empty `spec/fixtures` dir just in case?

Fryguy:
I don't really understand the spec/fixtures... I didn't think we used fixtures _at all_

jrafanie:
> Yeah I see a number of providers have `spec/fixtures`:
> 
> ```
> manageiq-providers-amazon/spec/fixtures:
> orchestration_templates
> 
> manageiq-providers-azure/spec/fixtures:
> events  orchestration_templates
> 
> manageiq-providers-azure_stack/spec/fixtures:
> orchestration_templates
> 
> manageiq-providers-kubevirt/spec/fixtures:
> files
> 
> manageiq-providers-openstack/spec/fixtures:
> orchestration_templates
> 
> manageiq-providers-vmware/spec/fixtures:
> orchestration_templates
> ```

These aren't failing on cross repo so I wonder if the problem is that this one uses a non-standard path: `"spec/fixtures/files"`

Others seem to use spec/fixtures/...

https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/spec/fixtures/orchestration_templates/cfn_parameters.json

And yet more just build then own path and don't use fixture_path at all: https://github.com/ManageIQ/manageiq-providers-vmware/blob/7e78afd5da4e05364156801408df06e57f56a896/spec/factories/orchestration_template.rb#L5

agrare:
> I didn't think we used fixtures at all

Yeah to be honest this is the first I've seen these

>  I wonder if the problem is that this one uses a non-standard path

Oh interesting

Issue #239, opened 2024-07-22
RefreshWorker fails to start due to nil parent_manager relation


agrare:
If the kubevirt provider is added to an OpenShift ContainerManager the parent_manager relation is `nil` since it specifies the class as `ManageIQ::Providers::Kubernetes::ContainerManager`

```
[----] E, [2024-07-22T14:57:12.663993#601005:a050] ERROR -- evm: MIQ(MiqServer::WorkerManagement::Process#sync_workers) Failed to sync_workers for class: ManageIQ::Providers::Kubevirt::InfraManager::RefreshWorker: undefined  method `where' for nil:NilClass 

    authentications.where(:authtype => :kubevirt)
                   ^^^^^^ 
[----] E, [2024-07-22T14:57:12.664698#601005:a050] ERROR -- evm: [NoMethodError]: undefined method `where' for nil:NilClass

    authentications.where(:authtype => :kubevirt)
                   ^^^^^^  Method:[block (2 levels) in <class:LogProxy>]
[----] E, [2024-07-22T14:57:12.664791#601005:a050] ERROR -- evm: /home/grare/adam/src/manageiq/manageiq-providers-kubevirt/app/models/manageiq/providers/kubevirt/infra_manager.rb:153:in `authentication_for_providers'
/home/grare/adam/src/manageiq/manageiq/app/models/mixins/authentication_mixin.rb:156:in `provider_authentication_status_ok?'
/home/grare/adam/src/manageiq/manageiq/app/models/mixins/provider_worker_mixin.rb:14:in `block in all_valid_ems_in_zone'
/home/grare/adam/.gem/ruby/3.1.0/gems/activerecord-6.1.7.8/lib/active_record/relation/delegation.rb:88:in `each' 
/home/grare/adam/.gem/ruby/3.1.0/gems/activerecord-6.1.7.8/lib/active_record/relation/delegation.rb:88:in `each' 
/home/grare/adam/.gem/ruby/3.1.0/gems/activerecord-6.1.7.8/lib/active_record/relation/query_methods.rb:269:in `select'
/home/grare/adam/.gem/ruby/3.1.0/gems/activerecord-6.1.7.8/lib/active_record/relation/query_methods.rb:269:in `select'
/home/grare/adam/src/manageiq/manageiq/app/models/mixins/provider_worker_mixin.rb:14:in `all_valid_ems_in_zone'
[...]

PR #240, opened 2024-07-23
Fix parent_manager class name

agrare:
The parent_manager class had Kubernetes::ContainerManager specified which would result in parent_manager being nil when kubevirt was added to an Openshift::ContainerManager due to ActAsStiLeafClass.

Since the base ContainerManager doesn't have an infra_manager we have to drop the inverse_of.

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/239


Fryguy:
Backported to radjabov via merge of master into radjabov branch

PR #242, opened 2024-07-25
Fix parser exception when memory value is missing

agrare:
```
[NoMethodError]: undefined method `/' for nil:NilClass
manageiq-providers-kubevirt/app/models/manageiq/providers/kubevirt/inventory/parser.rb:151:in `process_domain'
```


Fryguy:
Backported to radjabov via merge of master into radjabov branch

Issue #241, opened 2024-07-23
VM Provision Fails with 404 Not Found when getting the template


agrare:
```
[----] I, [2024-07-23T11:44:10.929024#74159:9c68]  INFO -- evm: MIQ(ManageIQ::Providers::Kubevirt::InfraManager#with_provider_connection) Connecting through ManageIQ::Providers::Kubevirt::InfraManager: [CRC Virtualization Manager]
[----] I, [2024-07-23T11:44:10.968173#74159:9c68]  INFO -- evm: Starting Phase <provision_error>
[----] E, [2024-07-23T11:44:10.974306#74159:9c68] ERROR -- evm: MIQ(ManageIQ::Providers::Kubevirt::InfraManager::Provision#provision_error) [[Fog::Kubevirt::Errors::NotFoundError]: HTTP status code 404, the server could not find the requested resource for GET https://api.crc.testing:6443/apis/template.openshift.io/v1/templates/rhel8-server-tiny] encountered during phase [start_clone_task]
[----] E, [2024-07-23T11:44:10.974422#74159:9c68] ERROR -- evm: /home/grare/adam/src/manageiq/fog-kubevirt/lib/fog/kubevirt/compute/utils/exception_wrapper.rb:26:in `rescue in method_missing'
/home/grare/adam/src/manageiq/fog-kubevirt/lib/fog/kubevirt/compute/utils/exception_wrapper.rb:11:in `method_missing'
/home/grare/adam/src/manageiq/fog-kubevirt/lib/fog/kubevirt/compute/requests/get_template.rb:6:in `get_template'
/home/grare/adam/src/manageiq/fog-kubevirt/lib/fog/kubevirt/compute/models/templates.rb:29:in `get'
/home/grare/adam/src/manageiq/manageiq-providers-kubevirt/app/models/manageiq/providers/kubevirt/infra_manager/connection.rb:80:in `template'
/home/grare/adam/src/manageiq/manageiq-providers-kubevirt/app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb:32:in `block in start_clone'
/home/grare/adam/src/manageiq/manageiq/app/models/ext_management_system.rb:568:in `with_provider_connection'
/home/grare/adam/src/manageiq/manageiq/app/models/mixins/provider_object_mixin.rb:5:in `with_provider_connection'
[...]

PR #249, opened 2024-09-11
Fix Kubevirt client get resource by name calls

nasark:
- `resource.get(name)` calls are namespace scoped
- since we use the catch-all namespace `""` when initializing the client https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/app/models/manageiq/providers/kubevirt/infra_manager.rb#L126, this will work for `resource.all` calls but will not work for `resource.get(name)` calls i.e. https://github.com/fog/fog-kubevirt/blob/master/lib/fog/kubevirt/compute/requests/get_template.rb#L6 and will return a 404

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/241

@miq-bot assign @agrare 
@miq-bot add_label bug, radjabov/yes?
@miq-bot add_reviewer @agrare

### DIFF
--- a/app/models/manageiq/providers/kubevirt/inventory/collector.rb
+++ b/app/models/manageiq/providers/kubevirt/inventory/collector.rb
@@ -27,7 +27,7 @@ class ManageIQ::Providers::Kubevirt::Inventory::Collector < ManageIQ::Providers:
     name = @target.name
     @nodes = {}
 
-    @manager.with_provider_connection do |connection|
+    @manager.with_provider_connection(:namespace => @target.location) do |connection|
       if @target.template?
         @templates = [connection.template(name)]
       else
--- a/app/models/manageiq/providers/kubevirt/infra_manager.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager.rb
@@ -184,7 +184,7 @@ class ManageIQ::Providers::Kubevirt::InfraManager < ManageIQ::Providers::InfraMa
       :host      => endpoint.hostname,
       :port      => endpoint.port,
       :token     => token,
-      :namespace => ""
+      :namespace => opts[:namespace] || ""
     )
   end
 
--- a/app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
@@ -28,7 +28,7 @@ module ManageIQ::Providers::Kubevirt::InfraManager::Provision::Cloning
   def start_clone(options)
     # Retrieve the details of the source template:
     template = nil
-    source.with_provider_connection do |connection|
+    source.with_provider_connection(:namespace => source.location) do |connection|
       template = connection.template(source.name)
       template.clone(user_options(options))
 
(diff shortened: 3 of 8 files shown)


agrare:
@nasark if you had two templates with the same name in two namespaces would this reliably find the right one?

nasark:
> @nasark if you had two templates with the same name in two namespaces would this reliably find the right one?

@agrare Good point, `.find` would return the first template that meets the criteria in the list which is not good. I guess another option is to add a provisioning option to specify the namespace of a template, then we wouldn't need these changes

agrare:
Yeah so since we've selected a template already we should know the namespace, is the ns part of the miq_template record at all?

agrare:
Interesting, if the template object from kubevirt has a namespace we should set that in the location just like vms do

app/models/manageiq/providers/kubevirt/infra_manager/connection.rb
```diff
@@ -17,18 +17,31 @@ class ManageIQ::Providers::Kubevirt::InfraManager::Connection
...
+    require 'fog/kubevirt'
+
+    namespaced_client = Fog::Kubevirt::Compute.new(
+      :kubevirt_hostname  => @opts[:host],
+      :kubevirt_port      => @opts[:port],
+      :kubevirt_token     => @opts[:token],
+      :kubevirt_namespace => namespace,
+      :kubevirt_log       => $log
+    )
+  end
```
nasark, line 43: `@conn` is the client connection that uses the catch-all namespace `""` and `namespaced_client` is the client connection targeting a specific namespace. Not a fan of this but didn't see another way until we proceed with replacing the fog client with `kubeclient`

app/models/manageiq/providers/kubevirt/infra_manager/connection.rb
```diff
@@ -77,7 +90,10 @@ def templates
   # @return [Object] The template object.
   #
   def template(name)
-    @conn.templates.get(name)
+    template = ManageIQ::Providers::Kubevirt::InfraManager::Template.find_by(:name => name)
+    client = namespaced_client(template.location)
+
+    client.templates.get(name)
```
agrare, line 96: Then we don't need to change these getters.  Side note `ManageIQ::Providers::Kubevirt::InfraManager::Template.find_by(:name => name)` isn't guaranteed to be correct since you could have multiple kubevirt providers and templates with the same name.

app/models/manageiq/providers/kubevirt/inventory/collector.rb
```diff
@@ -27,7 +27,7 @@ def initialize_for_targeted_refresh
     name = @target.name
     @nodes = {}
 
-    @manager.with_provider_connection do |connection|
+    @manager.with_provider_connection(:namespace => @target.location) do |connection|
       if @target.template?
```
agrare, line 31: Totally unrelated to this change, I don't think this can ever happen based on `if refresh_target.kind_of?(ManageIQ::Providers::Kubevirt::InfraManager::Vm)` above.  That wouldn't include `ManageIQ::Providers::Kubevirt::InfraManager::Template` instances

agrare (approved):
:+1: LGTM

Fryguy:
Backported to `radjabov` in commit faaee45d71c921a7f2bd97e8bcd0a2cd76405eb6.

```text
commit faaee45d71c921a7f2bd97e8bcd0a2cd76405eb6
Author: Adam Grare <adam@grare.com>
Date:   Mon Sep 23 15:10:26 2024 -0400

    Merge pull request #249 from nasark/kubevirt_service_catalog_provisioning
    
    Fix Kubevirt client get resource by name calls
    
    (cherry picked from commit 389a4937c4e835c608f11cff7568359415636b3b)
```

PR #263, opened 2024-12-10
Restart a Virtual Machine

Dhamo1107:
This PR introduces the "Restart VM" feature for VMs managed by the KubeVirt infrastructure provider in ManageIQ.

Testing:
Verified the restart functionality for KubeVirt VMs via the UI.

Before: "Restart Guest" option is disabled for KubeVirt VMs.

After: "Restart Guest" button available and functional for KubeVirt VMs.
image
image

@Fryguy Kindly take a look at this and review the PR. Thank you!

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
@@ -14,6 +14,10 @@ class ManageIQ::Providers::Kubevirt::InfraManager::Vm < ManageIQ::Providers::Inf
   supports :capture
   supports :snapshots
 
+  supports :reboot_guest do
+    _('The VM is not powered on') unless current_state == 'on'
+  end
+  supports :reset
   def self.calculate_power_state(raw)
     POWER_STATES[raw] || super
   end
--- a/app/models/manageiq/providers/kubevirt/infra_manager/vm/operations.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/vm/operations.rb
@@ -2,6 +2,7 @@ module ManageIQ::Providers::Kubevirt::InfraManager::Vm::Operations
   extend ActiveSupport::Concern
   include Power
   include Snapshot
+  include Guest
 
   included do
     supports(:terminate) { unsupported_reason(:control) }
--- a/app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/guest.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/guest.rb
@@ -2,13 +2,11 @@ module ManageIQ::Providers::Kubevirt::InfraManager::Vm::Operations::Guest
 
   def raw_reboot_guest
     kubevirt = ext_management_system.parent_manager.connect(:service => "kubernetes", :path => "/apis/subresources.kubevirt.io", :version => "v1")
-    #method need to call on this kubevirt to restart the vm
-
+    kubevirt.rest_client["namespaces/#{location}/virtualmachineinstances/#{name}/softreboot"].put({}.to_json, { 'Content-Type' => 'application/json' }.merge(kubevirt.get_headers))
   end
 
   def raw_reset
     kubevirt = ext_management_system.parent_manager.connect(:service => "kubernetes", :path => "/apis/subresources.kubevirt.io", :version => "v1")
-    #method need to call on this kubevirt to soft reboot the vm
-
+    kubevirt.rest_client["namespaces/#{location}/virtualmachines/#{name}/restart"].put({}.to_json, { 'Content-Type' => 'application/json' }.merge(kubevirt.get_headers))
   end
 end
--- a/app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/guest.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/guest.rb
@@ -1,0 +1,14 @@
+module ManageIQ::Providers::Kubevirt::InfraManager::Vm::Operations::Guest
+
+  def raw_reboot_guest
+    kubevirt = ext_management_system.parent_manager.connect(:service => "kubernetes", :path => "/apis/subresources.kubevirt.io", :version => "v1")
+    #method need to call on this kubevirt to restart the vm
+
+  end
+
+  def raw_reset
+    kubevirt = ext_management_system.parent_manager.connect(:service => "kubernetes", :path => "/apis/subresources.kubevirt.io", :version => "v1")
+    #method need to call on this kubevirt to soft reboot the vm
+
+  end
+end


app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
```diff
@@ -1,5 +1,6 @@
 class ManageIQ::Providers::Kubevirt::InfraManager::Vm < ManageIQ::Providers::InfraManager::Vm
   include Operations
+  include SupportsFeatureMixin
```
agrare, line 3: This is already included in the base class (https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L10) so you don't need to include it again
```suggestion
```

app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/power.rb
```diff
@@ -14,4 +14,12 @@ def raw_stop
       vm.stop
     end
   end
+
+  def raw_reboot_guest
+    ext_management_system.with_provider_connection(:namespace => location) do |connection|
+      # Retrieve the details of the virtual machine:
+      vm = connection.vm(name)
+      vm.restart
```
agrare, line 22: I need to see if `vm.restart` do a "soft" reboot via the guest or a restart via the virtualized hardware.  That would dictate if we call this reboot_guest or reset

agrare (commented):
:heart: thanks so much @Dhamo1107 !

app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/power.rb
agrare: Based on the API docs it looks like [`VirtualMachine#restart`](https://kubevirt.io/api-reference/v1.4.0/operations.html#_v1restart) is a hardware reset and [`VirtualMachineInstance#softreboot`](https://kubevirt.io/api-reference/v1.4.0/operations.html#_v1softreboot) is what we would call a reboot_guest

Dhamo1107: I added this in power.rb:
image
and i got this error:
image

nasark: I think you may need to override these in a subclass of https://github.com/ManageIQ/manageiq/blob/master/app/models/vm/operations/guest.rb

agrare: I forgot that `with_provider_connection` is still using fog-kubevirt since @nasark's most recent work here was using kubeclient.  I don't see restart or softreboot in fog-kubevirt so we might have an issue there.

Given that we want to move towards kubeclient in the near future as a replacement for fog-kubevirt I think we'd be better off going in that direction rather than adding restart/soft_reboot to fog-kubevirt.

Dhamo1107: Hi @agrare, As suggested I’ve connected to the VM using the following code:

`kubevirt = ext_management_system.parent_manager.connect(:service => "kubernetes", :path => "/apis/subresources.kubevirt.io", :version => "v1")  `

However, I’m not sure which method should be called on this kubevirt instance to perform the restart and reset operation (like `kubevirt.create_virtual_machine_snapshot`). I assume a similar operation should be implemented here using kubeclient. Can you please advise on the correct method I should use for this purpose?

agrare: Yeah so it looks like kubeclient doesn't pick up on the methods like reset and softreboot because the kubevirt.io API operations don't match the rest of the k8s apis.
I was able to restart and softreboot vms with:
```
client = osv.parent_manager.connect(:service => "kubernetes", :path => "/apis/subresources.kubevirt.io", :version => "v1")
client.rest_client['namespaces/manageiq/virtualmachines/fedora-jade-owl-56/restart'].put({}.to_json, { 'Content-Type' => 'application/json' }.merge(client.get_headers))
client.rest_client['namespaces/manageiq/virtualmachineinstances/fedora-jade-owl-56/softreboot'].put({}.to_json, { 'Content-Type' => 'application/json' }.merge(client.get_headers))
```

agrare:
Hey @Dhamo1107 FYI it looks like you have a merge commit in here, in our recommended git config https://www.manageiq.org/docs/guides/developer_setup/git_workflow we recommend `git config --global --bool pull.rebase true` so that you can `git pull upstream master` and it will rebase your commits onto master, rather than creating a new merge commit in your branch.  That just helps us keep the history a little cleaner.

Dhamo1107:
Hi @agrare, Got it, thanks for the input, I’ll update it as recommended

agrare:
I tested these changes and the reboot works perfectly, I saw a 500 error with soft_restart (I also hit this using kubeclient directly so it isn't anything in your PR that is causing it)

```
[----] E, [2024-12-16T10:26:20.612321#175292:97e0] ERROR -- evm: MIQ(MiqQueue#m_callback) Message id: [15721]: 500 Internal Server Error
[----] E, [2024-12-16T10:26:20.612535#175292:97e0] ERROR -- evm: [RestClient::InternalServerError]: 500 Internal Server Error  Method:[block (2 levels) in <class:LogProxy>]
[----] E, [2024-12-16T10:26:20.612641#175292:97e0] ERROR -- evm: /home/grare/adam/.gem/ruby/3.3.0/gems/rest-client-2.1.0/lib/restclient/abstract_response.rb:249:in `exception_with_response'
/home/grare/adam/.gem/ruby/3.3.0/gems/rest-client-2.1.0/lib/restclient/abstract_response.rb:129:in `return!'
/home/grare/adam/.gem/ruby/3.3.0/gems/rest-client-2.1.0/lib/restclient/request.rb:836:in `process_result'
/home/grare/adam/.gem/ruby/3.3.0/gems/rest-client-2.1.0/lib/restclient/request.rb:743:in `block in transmit'
/usr/lib/ruby/3.3.0/net/http.rb:1570:in `start'
/home/grare/adam/.gem/ruby/3.3.0/gems/rest-client-2.1.0/lib/restclient/request.rb:727:in `transmit'
/home/grare/adam/.gem/ruby/3.3.0/gems/rest-client-2.1.0/lib/restclient/request.rb:163:in `execute'
/home/grare/adam/.gem/ruby/3.3.0/gems/rest-client-2.1.0/lib/restclient/request.rb:63:in `execute'
/home/grare/adam/.gem/ruby/3.3.0/gems/rest-client-2.1.0/lib/restclient/resource.rb:79:in `put'
/home/grare/adam/src/manageiq/manageiq-providers-kubevirt/app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/guest.rb:5:in `raw_reboot_guest'
```

The VM does restart, it happens too quickly to see if the guest came down cleanly.  Possibly an issue with communicating with the guest tools?  Not sure...

agrare (approved):
Okay I fixed a merge commit and dropped a duplicate raw_reboot_guest
LGTM now, I'll have a follow-up PR to add a way to get a kubeclient instance without having to do `ext_management_system.parent_manager.connect(:service => "kubernetes", :path => "/apis/subresources.kubevirt.io", :version => "v1")`

Fryguy:
Backported to `radjabov` in commit f35ca3233a1b980b573fc56711e8b9f5682888a6.

```text
commit f35ca3233a1b980b573fc56711e8b9f5682888a6
Author: Adam Grare <adam@grare.com>
Date:   Thu Jan 2 09:02:24 2025 -0500

    Merge pull request #263 from Dhamo1107/vm_restart
    
    Restart a Virtual Machine
    
    (cherry picked from commit 4cb507a2d8bf2bca505d3bcdf6715aca70833fa9)
```

Issue #245, opened 2024-08-07
Investigate using Kubeclient directly rather than fog-kubevirt


agrare:
Since we don't really need the abstraction layer that Fog provides we could use the kubeclient sdk directly to hit the various kubevirt api groups.  This would give us more flexibility on staying up to date with the APIs and have one fewer dependency to keep up to date.

agrare:
https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/265

PR #265, opened 2024-12-10
[WIP] replace fog-kubevirt with kubeclient

agrare:
https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/245

TODO:
- [ ] Replace other callers of Connection with kubeclient calls


app/models/manageiq/providers/kubevirt/inventory/collector.rb
```diff
@@ -43,11 +43,26 @@ def initialize_for_targeted_refresh
...
-    @manager.with_provider_connection do |connection|
-      @nodes = connection.nodes
-      @vms = connection.vms
-      @vm_instances = connection.vm_instances
-      @templates = connection.templates
-    end
+    @nodes        = kube_client.get_nodes
+    @vms          = kubevirt_client.get_virtual_machines
+    @vm_instances = kubevirt_client.get_virtual_machine_instances
+    @templates    = openshift_template_client.get_templates
```
Fryguy, line 49: More out of curiosity, and not for this PR, but could we collect these from the clients in parallel?

PR #270, opened 2024-12-19
Kubevirt metrics capture specs

nasark:
We were missing some test coverage for Kubevirt metrics capture

@miq-bot assign @agrare 
@miq-bot add_label test
@miq-bot add_reviewer @agrare


app/models/manageiq/providers/kubevirt/infra_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -70,10 +70,10 @@ def labels_to_s(labels)
   end
 
   def validate_target
-    raise TargetValidationError, "ems not defined" unless @ext_management_system
-    raise TargetValidationWarning, "no associated hardware" unless @vm_hardware
+    raise ManageIQ::Providers::Kubevirt::InfraManager::MetricsCapture::TargetValidationError, "ems not defined" unless @ext_management_system
+    raise ManageIQ::Providers::Kubevirt::InfraManager::MetricsCapture::TargetValidationWarning, "no associated hardware" unless @vm_hardware
 
-    raise TargetValidationError, "cores not defined" unless @vm_cores.to_i > 0
-    raise TargetValidationError, "memory not defined" unless @vm_memory.to_i > 0
```
nasark, line 77: This was needed since the capture context is subclassed from k8s which has its own error handling classes
```
class ManageIQ::Providers::Kubevirt::InfraManager::MetricsCapture::PrometheusCaptureContext < ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::PrometheusCaptureContext
```

This should not be needed if https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/543 is resolved but that may necessitate other changes to the specs

config/secrets.defaults.yml
```diff
@@ -0,0 +1,8 @@
+---
+test:
+  kubevirt_defaults: &kubevirt_defaults
+    hostname: KUBEVIRT_HOSTNAME
+    token: KUBEVIRT_TOKEN
+    metrics_hostname: METRICS_HOSTNAME
+  kubevirt:
+    <<: *kubevirt_defaults
```
Fryguy, line 8: I can't remember if we changed this yet or not.  cc @jrafanie

jrafanie: Not yet.  I haven't had time to get back to it.  The referenced changes are coming from this core change: https://github.com/ManageIQ/manageiq/pull/23292 and how each plugin would use these changes: https://github.com/ManageIQ/manageiq-providers-vmware/pull/928

jrafanie: I might just leave the warnings with rails 7.1 for now and people can fix them when their usage of rails secrets in 7.1 doesn't work or gets a warning.  Note, I don't think we actually use rails to store secrets encrypted.  That's the part that's completely gone in rails 7.1.  Access is probably gone in 7.2.

Issue #268, opened 2024-12-13
Add Console Access for KubeVirt VMs


Dhamo1107:
ManageIQ currently does not support console access for KubeVirt VMs, a feature that is available for OpenStack instances. Enabling this functionality would allow users to interact directly with their KubeVirt VMs for troubleshooting, configuration and recovery. Please provide your suggestions on how to implement this feature. Thank you for your consideration!

Dhamo1107:
Hi @agrare,

From the [Readme](https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/3b3d098a5fb5ba7ce12db2a259766270847f4f95/README.md?plain=1#L21) of the KubeVirt provider, it appears that we can connect to the console of VMs (VNC is used in our OpenShift environment). However, on the VM detail page, the `Access -> VM Console` option is currently not available.

In this WIP [PR](https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/269) I have added the `RemoteConsole` module. 

As a result, the VM Console option is now visible.
![image](https://github.com/user-attachments/assets/e2faa4ef-7db9-45ed-9e5d-659ec801cf87)
![image](https://github.com/user-attachments/assets/c2fdd440-ba88-49e0-a9ac-200e8f62d102)

How should I proceed further from here? Please provide your insights on the next steps or any improvements required.

Thank You!

agrare:
It looks like there is an API to get a web socket for a VNC console, https://kubevirt.io/api-reference/v1.4.0/operations.html#_v1vnc

And the web socket template is `wss://%s:%v/v1/namespaces/%s/virtualmachineinstances/%s/vnc` (based on the virtctl code)

agrare:
You'll need to implement the `remote_console_acquire_ticket_queue` and `remote_console_acquire_ticket` methods.  The OpenStack VM VNC model is probably a good one to follow here, https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/cloud_manager/vm/remote_console.rb

Dhamo1107:
Hi @agrare, 

I used this API: [KubeVirt v1.4.0 API Reference](https://kubevirt.io/api-reference/v1.4.0/operations.html#_v1vnc)  

I have the following method:  

```ruby
def remote_console_acquire_ticket(_userid, _originating_server, _console_type)
  url = ext_management_system.with_provider_connection(:namespace => location) do |connection|
[...]

PR #272, opened 2025-01-07
Remove obsolete SPICE console option

agrare:
This was removed by kubevirt in https://github.com/kubevirt/kubevirt/pull/618 and replaced with VNC support

Related: https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/268 adding VNC support

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
@@ -23,68 +23,6 @@ class ManageIQ::Providers::Kubevirt::InfraManager::Vm < ManageIQ::Providers::Inf
   end
 
   #
-  # This method is executed in the UI worker. It adds to the queue a task to retrieve the
-  # `remote-viewer` configuration file for a virtual machine.
-  #
-  # @return [Integer] The identifier of the task.
-  #
-  def queue_generate_remote_viewer_file
-    # Create the task options:
-    task_options = {
-      :action => "Generate 'remote-viewer' configuration file for VM '#{name}'"
-    }
-
-    # Create the queue options:
-    queue_options = {
-      :class_name  => self.class.name,
-      :instance_id => id,
-      :method_name => 'generate_remote_viewer_file',
-      :priority    => MiqQueue::HIGH_PRIORITY,
-      :role        => 'ems_operations',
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
```diff
@@ -49,41 +49,6 @@ def queue_generate_remote_viewer_file
     MiqTask.generic_action_with_callback(task_options, queue_options)
   end
 
-  #
-  # This method is executed in the provider worker. It contacts the KubeVirt API to get the details of the connection to
-  # the console, and generates the `remote-viewer` configuration file.
-  #
-  # @return [String] The content of the cofiguration file.
-  #
-  def generate_remote_viewer_file
```
agrare, line 58: As far as I can tell this was never called

Fryguy: Should this be killed also, then?

https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/4cb507a2d8bf2bca505d3bcdf6715aca70833fa9/app/models/manageiq/providers/kubevirt/infra_manager/vm.rb#L31

Fryguy: I also found this semi-related method in ovirt (and rhv):

https://github.com/ManageIQ/manageiq-providers-ovirt/blob/8c8a8e3d01c2b1e766bc690a076e6cca22416d1b/app/models/manageiq/providers/ovirt/infra_manager/ovirt_services.rb#L205

agrare: > Should this be killed also, then?

Yes, done

> I also found this semi-related method in ovirt (and rhv):

Ovirt/RHV still use spice so that is still used, spice was only removed from kubevirt

Fryguy: Weird - the comment on the queue method says ui-classic calls it - but I checked the entire diff history of ui-classic and it's not there.

agrare: Yeah I checked the same, my guess is that it was started but never finished

PR #276, opened 2025-02-18
Replace fog-kubevirt with kubevirt-sdk-ruby

agrare:
Alternative to https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/265
This uses an openapi generated client gem (https://github.com/ManageIQ/kubevirt-sdk-ruby/) based on https://github.com/kubevirt/kubevirt/tree/main/api/openapi-spec

Because the kubevirt API doesn't follow the same rules as most of the rest of the k8s API, kubeclient isn't able to generate methods for all endpoints.  This is also true of the `oc` utility which is why there is a `virtctl` cli (https://kubevirt.io/user-guide/user_workloads/virtctl_client_tool/)


Fryguy:
Nice - I assume we'll want to publish kubevirt-sdk-ruby under the ManageIQ org first?

agrare:
Yes, still a good amount of work to do to move over to using the new gem so no rush on transfer

app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -122,13 +123,20 @@ def self.verify_credentials(args)
...
+    require "kubevirt"
+
+    api_key = ManageIQ::Password.try_decrypt(opts[:token])
+
+    kubevirt_api_client = Kubevirt::ApiClient.new
+    kubevirt_api_client.config.api_key    = api_key
+    kubevirt_api_client.config.scheme     = "https"
+    kubevirt_api_client.config.host       = "#{opts[:server]}:#{opts[:port] || 6443}"
+    kubevirt_api_client.config.verify_ssl = false # TODO check security_protocol
+    kubevirt_api_client.config.logger     = $log # TODO add a kubevirt logger
```
Fryguy, line 135: I would expect the logger to come in from the opts otherwise we should probably default to a null logger (or just not set it)

app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -122,13 +123,20 @@ def self.verify_credentials(args)
...
+    kubevirt_api_client.config.debugging  = Settings.log.level == "debug"
+    kubevirt_api_client.default_headers["Authorization"] = "Bearer #{api_key}"
```
Fryguy, line 137: Modifying the default headers seems weird to me? Is that just how the interface works?

app/models/manageiq/providers/kubevirt/infra_manager.rb
agrare: Yeah I just copied the use of `$log` from the existing Connection class, but per the note we should add a kubevirt logger

app/models/manageiq/providers/kubevirt/infra_manager.rb
agrare: Agree it was weird to me also, it seems like an openapi thing since we do the same in powervs https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb#L27

I think if the openapi spec declared an auth setting then the openapi generator would handle this for you, so far this was the only way to get the openapi generated gem to send authorization headers.

app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/guest.rb
```diff
@@ -1,12 +1,14 @@
 module ManageIQ::Providers::Kubevirt::InfraManager::Vm::Operations::Guest
 
   def raw_reboot_guest
-    kubevirt = ext_management_system.parent_manager.connect(:service => "kubernetes", :path => "/apis/subresources.kubevirt.io", :version => "v1")
-    kubevirt.rest_client["namespaces/#{location}/virtualmachineinstances/#{name}/softreboot"].put({}.to_json, { 'Content-Type' => 'application/json' }.merge(kubevirt.get_headers))
```
agrare, line 5: This was a good example of a method call that kubeclient couldn't handle, and `oc vm` can't do either and you have to use virtctl for.  (This was always gross but didn't want to block @Dhamo1107 from adding support for this while he was waiting for me to do this change).

app/models/manageiq/providers/kubevirt/infra_manager.rb
agrare: https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/279

app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -109,7 +109,8 @@ def self.params_for_create
 
   def self.verify_credentials(args)
     kubevirt = raw_connect(args.dig("endpoints", "default")&.slice("server", "port", "token")&.symbolize_keys)
-    kubevirt&.valid? && kubevirt&.virt_supported?
+    kubevirt&.get_api_group_list
+    kubevirt&.get_api_group_kubevirt_io
```
agrare, line 113: TODO I need to confirm this behavior on an cluster that doesn't have kubevirt/osv installed

agrare: image

agrare:
@Fryguy okay I think we're at the point where we can start moving kubevirt-sdk-ruby to the ManageIQ org and doing a gem release

Fryguy:
@agrare https://github.com/ManageIQ/kubevirt-sdk-ruby is ready to go.  There might be some minor cleanup, then you can cut the gem.  Please include me as a backup owner when you do.

app/models/manageiq/providers/kubevirt/infra_manager.rb
Fryguy: Are we still waiting on this verification?

app/models/manageiq/providers/kubevirt/infra_manager.rb
```diff
@@ -150,7 +168,17 @@ def has_credentials?(_type = nil)
   # @param opts [Hash] Additional options to control how to perform the verification.
   #
   def verify_credentials(_type = nil, opts = {})
-    with_provider_connection(opts, &:valid?)
+    with_provider_connection(opts) do |kubevirt|
+      # First check that we're able to connect to the k8s API
+      kubevirt.get_api_group_list
+      # Then check that we can access the /apis/kubevirt.io API group
+      kubevirt.get_api_group_kubevirt_io
```
Fryguy, line 175: Since we're doing this in 2 spots, I wonder if we should move this into a method.

Fryguy: Actually is this a duiplicate method in the same file?

app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
```diff
@@ -54,4 +78,60 @@ def user_options(options)
...
+    default_params
+  end
+
+  def param_substitution!(object, params)
+    result = object
+    case result
+    when Hash
+      result.each do |k, v|
+        result[k] = param_substitution!(v, params)
+      end
```
Fryguy, line 102: Can this be 

```suggestion
      result.transform_values! { |v| param_substitution!(v, params) }
```

app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
```diff
@@ -54,4 +78,60 @@ def user_options(options)
...
+  end
+
+  #
+  # Performs substitution on specific object.
+  #
+  # @params params [Hash] Containing parameter names and values used for substitution.
+  # @params object [String] Object on which substitution takes place.
+  # @returns [String] The outcome of substitution.
+  #
+  def sub_specific_object(params, object)
```
Fryguy, line 118: Method name is weird to me - thinking `substitute_string` or `substitute_literal`.  Not important for merge.

app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
```diff
@@ -26,18 +25,43 @@ def log_clone_options(clone_options)
...
+      namespace ||= source.location
 
-      vm = connection.vm(options[:name])
+      params    = values(template, user_options(options))
+      pvc_specs = template.objects.select { |o| o[:kind] == "PersistentVolumeClaim" }&.map(&:to_h)
+      vm_spec   = template.objects.detect { |o| o[:kind] == "VirtualMachine" }&.to_h
+      raise N_("No Virtual Machine defined in template") if vm_spec.nil?
 
-      phase_context[:new_vm_ems_ref] = vm.uid
+      param_substitution!(vm_spec, params)
```
Fryguy, line 38: Is this a new feature? I don't see it in any of the deleted code, and it didn't do any of this previously.

app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
agrare: This came from fog-kubevirt https://github.com/fog/fog-kubevirt/blob/master/lib/fog/kubevirt/compute/models/template.rb#L189-L206

app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
```diff
@@ -54,4 +78,60 @@ def user_options(options)
...
+    template.parameters.each do |param|
+      name = param[:name].downcase
+      value = options[name.to_sym]
+      if value && name == "memory"
+        value = "#{value}Mi"
+      end
+
+      default_params[name] = value || param[:value]
+    end
+    default_params
```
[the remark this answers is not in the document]
agrare: This was another method from fog-kubevirt, https://github.com/fog/fog-kubevirt/blob/master/lib/fog/kubevirt/compute/models/template.rb#L59-L71

In general I tried to not change these too much other than really obvious stuff so we weren't changing too many things at once

app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
agrare: Sure

Fryguy: Oh I didn't realize this came from elsewhere - it can be changed separately.

Fryguy:
Ok, don't worry about the copied code - that can all be done separately.

app/models/manageiq/providers/kubevirt/infra_manager.rb
agrare: No, above screenshot shows it working

PR #281, opened 2025-02-26
Fix alignment of delegate keys

agrare:


Fryguy:
marking for backport just to avoid any possible future conflicts.

PR #283, opened 2025-03-06
Parse host memory and cpu

agrare:
Built on:
- [x] https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/276

### DIFF
--- a/app/models/manageiq/providers/kubevirt/inventory/parser.rb
+++ b/app/models/manageiq/providers/kubevirt/inventory/parser.rb
@@ -22,6 +22,7 @@ class ManageIQ::Providers::Kubevirt::Inventory::Parser < ManageIQ::Providers::In
   attr_reader :cluster_collection
   attr_reader :host_collection
   attr_reader :host_storage_collection
+  attr_reader :host_hw_collection
   attr_reader :hw_collection
   attr_reader :network_collection
   attr_reader :os_collection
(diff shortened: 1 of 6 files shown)


app/models/manageiq/providers/kubevirt/inventory/parser.rb
```diff
@@ -82,6 +83,18 @@ def process_node(object)
...
+    hw_object = host_hw_collection.find_or_build(host_object)
+
+    memory = object.status&.capacity&.memory
+    memory &&= begin
+      memory.iec_60027_2_to_i
+    rescue
+      memory.decimal_si_to_f
+    end
+
+    hw_object.memory_mb = memory / 1.megabyte if memory
```
Fryguy, line 95: I can't tell, but is it possible for this to end up with integer division if memory is an integer?  Should we force float division here with something like?

```suggestion
    hw_object.memory_mb = memory / 1.megabyte.to_f if memory
```

spec/models/manageiq/providers/kubevirt/infra_manager/refresher_spec.rb
```diff
@@ -67,6 +67,10 @@ def assert_specific_host
         :vmm_version      => "0.1.0",
         :ems_cluster      => ems.ems_clusters.find_by(:ems_ref => "0")
       )
+      expect(host.hardware).to have_attributes(
+        :cpu_total_cores => 8,
+        :memory_mb       => 24_024
```
Fryguy, line 72: For example, here - should this be a float?

app/models/manageiq/providers/kubevirt/inventory/parser.rb
Fryguy: Ok then I have the opposite concern in that memory could be a float from `decimal_si_to_f`

agrare: Pretty sure AR assignment to a bigint column would floor that for us, interestingly this is what we do in the VMware parser `(memory_size.to_f / 1.megabyte).round`, also this specific logic above was ~blatantly ripped off~ "borrowed" from the kubernetes parser.  If you want we could `(memory.to_f / 1.megabyte).floor` here and in kubernetes but I'm not sure if you can have a fractional `Mi` value in k8s (easy to test....)

Fryguy: yeah I think we should be explicit, so the .to_f .floor seems appropriate

Fryguy (approved):

PR #284, opened 2025-03-06
VM Clone Cleanups

agrare:
Follow-up to https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/276

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
@@ -96,7 +96,7 @@ module ManageIQ::Providers::Kubevirt::InfraManager::Provision::Cloning
     when Array
       object.map! { |v| param_substitution!(v, params) }
     when String
-      object = sub_specific_object(params, object)
+      object = substitute_string_param(object, params)
     end
   end
 
@@ -107,7 +107,7 @@ module ManageIQ::Providers::Kubevirt::InfraManager::Provision::Cloning
   # @params object [String] Object on which substitution takes place.
   # @returns [String] The outcome of substitution.
   #
-  def sub_specific_object(params, object)
+  def substitute_string_param(object, params)
     result = object
     params.each_key do |name|
       token = "${#{name.upcase}}"
--- a/app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
@@ -96,34 +96,29 @@ module ManageIQ::Providers::Kubevirt::InfraManager::Provision::Cloning
     when Array
       object.map! { |v| param_substitution!(v, params) }
     when String
-      object = substitute_string_param(object, params)
+      substitute_string_param!(object, params)
     end
   end
 
   #
   # Performs substitution on specific object.
   #
-  # @params params [Hash] Containing parameter names and values used for substitution.
   # @params object [String] Object on which substitution takes place.
+  # @params params [Hash] Containing parameter names and values used for substitution.
   # @returns [String] The outcome of substitution.
   #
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
```diff
@@ -115,7 +107,7 @@ def param_substitution!(object, params)
   # @params object [String] Object on which substitution takes place.
   # @returns [String] The outcome of substitution.
   #
-  def sub_specific_object(params, object)
+  def substitute_string_param(object, params)
```
Fryguy, line 110: Can we name this with a `!` since it modifies the parameter?  Also, interestingly, it says `result = object`, then later does object.sub!, but takes the return value and puts it back in result.

agrare: Yeah I'm not sure this was ever working in the case where the param is not a string...working on a spec test for that case

agrare: Okay done, I feel better changing this method now that it has better spec coverage :laughing:

spec/models/manageiq/providers/kubevirt/infra_manager/provision/cloning_spec.rb
```diff
@@ -42,6 +47,22 @@
       expect(subject.phase_context[:new_vm_ems_ref]).to eq(new_vm.metadata.uid)
     end
 
+    context "with hash parameters" do
+      let(:options)    { {:cores_per_socket => {:cores => 2, :sockets => 2, :threads => 4}, :vm_memory => ["1024", "1024"], :root_password => "1234"} }
```
agrare, line 51: NOTE this `:cores_per_socket => {:cores => 2, :sockets => 2, :threads => 4}` and `:spec => hash_including(:domain => {:cores => 2, :sockets => 2, :threads => 4}` tests the non-string case of the `substitute_string_param!` method https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/284/files#diff-4da493788b2d0b8f8e9969e0b7dd53c65c1c0dbe421837be10467e81a5849c13R119

app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
```diff
@@ -80,58 +80,45 @@ def user_options(options)
...
   # Performs substitution on specific object.
   #
-  # @params params [Hash] Containing parameter names and values used for substitution.
   # @params object [String] Object on which substitution takes place.
+  # @params params [Hash] Containing parameter names and values used for substitution.
   # @returns [String] The outcome of substitution.
   #
-  def sub_specific_object(params, object)
-    result = object
+  def substitute_string_param!(object, params)
```
agrare, line 110: I don't think substitute _string_ param is a good name though since this does more than just strings

Fryguy: ah ok - glad you dug in - from the original code it felt like it was only called in the string case in the case statement, so that's why I suggested string in the name.  Up to you what to call it.

Fryguy:
I'm good to merge, unless you want to change that method name

app/models/manageiq/providers/kubevirt/infra_manager/provision/cloning.rb
agrare: Yeah the String case is when the value from the object that we're substituting is a string like: `:spec => {:domain => "${CPU}"}` it is the "${CPU}" case.  Then in substitute_string_param it is looking at what is being substituted.

agrare:
Okay how about `substitute_object_value!` since it is replacing the value in the object :) naming is hard

Fryguy (approved):

PR #285, opened 2025-03-06
VM Reconfigure

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/283
- [x] https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/287

TODO:
- [x] Specs
- [x] Host parser has to set cpus/memory
- [x] Sockets vs total cpus (kubevirt lets you do either, just set num cpus or set topology manually and core assumes you'll set manually, can we even get sockets from kubevirt nodes?)
- [x] What to do if host is nil (vmware would set this based on hardware version but maybe we set it based on the most cpus on any kubevirt host?)

### DIFF
--- a/spec/models/manageiq/providers/kubevirt/infra_manager/vm/reconfigure_spec.rb
+++ b/spec/models/manageiq/providers/kubevirt/infra_manager/vm/reconfigure_spec.rb
@@ -1,0 +1,72 @@
+describe ManageIQ::Providers::Kubevirt::InfraManager::Vm::Reconfigure do
+  let(:ems) { FactoryBot.create(:ems_kubevirt) }
+  let(:vm) { FactoryBot.create(:vm_kubevirt, :ext_management_system => ems) }
+  let(:miq_server) { EvmSpecHelper.local_miq_server }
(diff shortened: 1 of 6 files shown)


agrare:
Hey @Fryguy can you review?

spec/models/manageiq/providers/kubevirt/infra_manager/vm/reconfigure_spec.rb
```diff
@@ -0,0 +1,102 @@
+describe ManageIQ::Providers::Kubevirt::InfraManager::Vm::Reconfigure do
+  let(:ems) { FactoryBot.create(:ems_kubevirt) }
+  let(:vm) { FactoryBot.create(:vm_kubevirt, :ext_management_system => ems) }
+  let(:miq_server) { EvmSpecHelper.local_miq_server }
+
+  describe "reconfigurable?" do
```
Fryguy, line 6: super-minor

```suggestion
  describe "#reconfigurable?" do
```

Issue #286, opened 2025-03-11
Support parsing VMs from InstanceTypes


agrare:
Currently we support parsing vm cpu/memory from the spec.template property but if a vm is created from an instance_type this information isn't available.

![Image](https://github.com/user-attachments/assets/c225e506-a3ca-403a-b5c9-74a3277bc33e)

agrare:
NOTE this will be required for supporting reconfigure of VMs created from instance types

PR #288, opened 2025-03-13
[WIP] Collect kubevirt instance types

agrare:
There are two options for creating a VM in Kubevirt, one is via a Template the other is from an InstanceType.  If created from an InstanceType there is no spec.template property which is how we currently get cpu/memory.

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/286


Fryguy:
@agrare Is this ready to merge?

Fryguy:
Can we add an assert_specific_flavor test?

PR #291, opened 2025-03-31
[WIP] More standard inventory/{collector,parser,persister}

agrare:
While writing specs for a streaming/watch notice based refresh I noticed that
1. The streaming refresh was going down the "initialize_for_full_refresh" branch in the collector because the target wasn't a vm
2. VM targeted refresh doesn't actually pass a `::Vm` object so it was _also_ doing a full

The KubeVirt inventory/* layout is unique :tm: and not like any other provider which makes it more difficult to maintain.
Since KubeVirt refresh is essentially the same as Kubernetes/OpenShift this moves towards that style

Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/292


Fryguy:
Merged #292 , so you can rebase this.

PR #292, opened 2025-03-31
Cleanup unused VM Targeted Refresh Code

agrare:
The current VM targeted refresh code is unreachable because:
1. Targets are converted into an `InventoryRefresh::TargetCollection`
2. `Kubevirt::InfraManager#allow_targeted_refresh?` is false so core converts to a full refresh anyway

Targeting a specific VM isn't necessary since this provider uses watches to get updates in near-real-time


Fryguy (approved):

agrare:
@jrafanie  https://github.com/ManageIQ/manageiq-providers-openshift/pull/282

Issue #296, opened 2025-06-09
Reconfigure of a VM created from an instance_type fails with HTTP 422 error


agrare:
When reconfiguring a VM that was created from an instance_type rather than a template we are setting values which are not applicable to that type of VM which causes kubevirt/osv to throw an HTTP 422 error code.

Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/286

PR #299, opened 2025-07-09
Allow VM reconfigure with instance_types / flavors

agrare:
If a VM is created from an instance_type it doesn't have a `spec/template/spec/domain/{cpu,memory}` section but rather a `spec/instancetype` section.

We still need to change the reconfigure dialog to optionally display instance_types instead of just cpu/memory but this gets the API call to kubernetes working.

https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/296


Fryguy (approved):

agrare:
> I'm fine not doing this rubocop for internal consistency.

Yeah I thought it read a little better this way

PR #301, opened 2025-07-18
Volume attach detach

chitra607:


agrare:
Hey @chitra607 is there a reason you went with `kubeclient` instead of `kubevirt-sdk` ?  I see `v1vm_addvolume` https://github.com/ManageIQ/kubevirt-sdk-ruby/blob/master/lib/kubevirt/api/default_api.rb#L16324-L16333

agrare:
@chitra607 interesting, we do use kubevirt-sdk-ruby's `patch_namespaced_virtual_machine` method which looks like what you're using here:
https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/app/models/manageiq/providers/kubevirt/infra_manager/vm/operations/configuration.rb

agrare:
Can you rebase your commits so we don't have a number of `updated` commits?

chitra607:
> Can you rebase your commits so we don't have a number of `updated` commits?

Sure, I’ll rebase the commits again and push the updated branch

chitra607:
updated branch 
https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/302/commits/0b21b9cea2fce324e2b5ccebf3b065ff48dc08e2

Issue #306, opened 2025-10-27
VM Provision Fails with error "Running and RunStrategy are mutually exclusive."


agrare:
```
{
  "kind"=>"Status",
  "apiVersion"=>"v1",
  "metadata"=>{},
  "status"=>"Failure",
  "message"=>"admission webhook \"virtualmachine-validator.kubevirt.io\" denied the request: Running and RunStrategy are mutually exclusive. Note that Running is deprecated, please use RunStrategy instead",
  "reason"=>"Invalid",
  "details"=>{
    "causes"=>[
      {
        "reason"=>"FieldValueInvalid",
        "message"=>"Running and RunStrategy are mutually exclusive. Note that Running is deprecated, please use RunStrategy instead",
        "field"=>"spec.running"
      }
    ]
  },
  "code"=>422
}

PR #308, opened 2025-10-27
Fix vm spec running run strategy

agrare:
Fix spec.running and runStrategy both set for a VM The `v1.VirtualMachineSpec` has deprecated `running` in favor of
`runStrategy` and will fail if both are present.

This fixes the error `Running and RunStrategy are mutually exclusive. Note that Running is deprecated, please use RunStrategy instead`

https://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachinespec

Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/307

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/306


Fryguy:
Merged #307

PR #310, opened 2025-10-29
Cleanup unreachable targeted refresh code and split full/partial inventory collector

agrare:
Kubevirt uses streaming refresh and does not support classic "targeted" refresh.  Even if you did request a targeted refresh of a single VM the refresh_targets are converted into a `InventoryRefresh::TargetCollection` and not a `Vm` instance like this was expecting.

### DIFF
--- a/app/models/manageiq/providers/kubevirt/inventory/collector.rb
+++ b/app/models/manageiq/providers/kubevirt/inventory/collector.rb
@@ -15,40 +15,10 @@ class ManageIQ::Providers::Kubevirt::Inventory::Collector < ManageIQ::Providers:
   def initialize(manager, refresh_target)
     super
 
-    if refresh_target.kind_of?(ManageIQ::Providers::Kubevirt::InfraManager::Vm)
-      initialize_for_targeted_refresh
-    else
-      initialize_for_full_refresh
-    end
-  end
-
-  protected
-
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubevirt/inventory/collector.rb
```diff
@@ -15,40 +15,10 @@ class ManageIQ::Providers::Kubevirt::Inventory::Collector < ManageIQ::Providers:
   def initialize(manager, refresh_target)
     super
 
-    if refresh_target.kind_of?(ManageIQ::Providers::Kubevirt::InfraManager::Vm)
-      initialize_for_targeted_refresh
-    else
-      initialize_for_full_refresh
-    end
```
agrare: NOTE there's a bug here which is fixed in another PR (https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/311), but https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/app/models/manageiq/providers/kubevirt/infra_manager/refresh_worker/runner.rb#L164 passes in `nil` when building the targeted collector so it actually fetches full inventory first, then replaces it with the payload from the watch notices.

This appears to have been a bug for many many years unfortunately

Fryguy (approved):

PR #311, opened 2025-10-29
Fix full refresh collection during targeted

agrare:
The `initialize_for_full_refresh` method was being called for targeted refreshes because we are passing `nil` in for the `refresh_target` [[ref]](https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/app/models/manageiq/providers/kubevirt/infra_manager/refresh_worker/runner.rb#L164) which went down the full_refresh conditional branch in the collector initializer [[ref]](https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/app/models/manageiq/providers/kubevirt/inventory/collector.rb#L18-L22)

Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/310

Follow-up:
* https://github.com/ManageIQ/manageiq-providers-openshift/pull/293

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager/refresh_worker/runner.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/refresh_worker/runner.rb
@@ -61,8 +61,8 @@ class ManageIQ::Providers::Kubevirt::InfraManager::RefreshWorker::Runner < Manag
     provider_class::InfraManager
   end
 
-  def collector_class
-    provider_class::Inventory::Collector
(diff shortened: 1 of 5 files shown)


agrare:
NOTE this doesn't follow any of the other conventions for full/targeted or even full+watches from k8s/ocp.

I tried to see if moving to that could be done quickly but the "weirdness" is too spread out so I'm going to do it in batches, and just focus on fixing this bug by itself first.

Fryguy (approved):

PR #327, opened 2026-08-04
Add HTML5 VNC remote console support for KubeVirt VMs

ishaaq-48:
# Purpose

Add support for acquiring HTML5 VNC remote console tickets for KubeVirt virtual machines.

This implementation introduces the KubeVirt-specific remote console acquisition flow and creates the required SystemConsole record for proxying console connections through ManageIQ.

## Changes

- Add KubeVirt remote console support.
- Implement VNC ticket acquisition.
- Add HTML5 console support using the existing VNC workflow.
- Create SystemConsole entries for KubeVirt consoles.
- Queue remote console acquisition through EMS operations.

## Testing

- Verified remote console ticket generation.
- Verified SystemConsole records are created.
- Verified console launch request reaches ManageIQ.

## Depends On

- https://github.com/ManageIQ/manageiq-ui-classic/pull/10204
- https://github.com/ManageIQ/manageiq/pull/23944

### DIFF
--- a/app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
@@ -1,7 +1,8 @@
 class ManageIQ::Providers::Kubevirt::InfraManager::Vm < ManageIQ::Providers::InfraManager::Vm
   include Operations
   include Reconfigure
-
+  include RemoteConsole
+  
   POWER_STATES = {
     'Running'    => 'on',
     'Pending'    => 'powering_up',
--- a/app/models/manageiq/providers/kubevirt/infra_manager/vm/remote_console.rb
+++ b/app/models/manageiq/providers/kubevirt/infra_manager/vm/remote_console.rb
@@ -1,0 +1,52 @@
+module ManageIQ::Providers::Kubevirt::InfraManager::Vm::RemoteConsole
+  def console_supported?(type)
+    %w[VNC HTML5].include?(type.upcase)
+  end
+
+  def validate_remote_console_acquire_ticket(protocol, _options = {})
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/kubevirt/infra_manager/vm.rb
```diff
@@ -1,7 +1,8 @@
 class ManageIQ::Providers::Kubevirt::InfraManager::Vm < ManageIQ::Providers::InfraManager::Vm
   include Operations
   include Reconfigure
-
+  include RemoteConsole
+
```
agrare:
```suggestion

```

app/models/manageiq/providers/kubevirt/infra_manager/vm/remote_console.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  def kubevirt_namespace
+    location.presence.tap do |ns|
+      if ns.blank? || ns == "unknown"
+        raise MiqException::RemoteConsoleNotSupportedError,
+              "Unable to determine the namespace for VM #{name}. A provider refresh may be required."
+      end
+    end
+  end
```
agrare: I think this would be better as a `validate_vm_namespace` method up [here](https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/327/changes#diff-053a5e67cea0268f2e4db4388b15031af9069d40d2d431356b9db710fc238988R44) instead of as an accessor method that sometimes raises an exception.


```suggestion
  def validate_vm_namespace
    if location.blank? || location == "unknown"
      raise MiqException::RemoteConsoleNotSupportedError,
              "Unable to determine the namespace for VM #{name}. A provider refresh may be required."
    end
  end
```

and then just use `location` in the `kubevirt_vnc_path` path

ishaaq-48: yes, I will update it.

